PR #30, opened 2025-09-01
[valkey] Add support for serviceAccount

### LINKED ISSUE #29: Charts are missing serviceAccounts
### Name and Version

valkey

It looks like all charts are currently just using the default serviceAccount. Best practice is to create a serviceAccount for every Deployment for easier RBAC management.

I'd be happy to contirbute here. Would just add a serviceAccount template for all charts, enabled by default.

_No response_

### Additional informations

_No response_

matt-mai:
### Description of the change

Adds serviceAccount creation by default with option to disable. 
This is standard with `helm create` as well.

### Benefits

Allow configuring serviceAccount instead of using default sa. 
Also lets users set automountServiceAccountToken (false, by default)

### Possible drawbacks

serviceAccount generation is enabled by default, which would lead to restart for existing deployments. 
We could instead disable by default, but that would lead to worse security on new default installs, as using the default sa with token mounted is discouraged.

### Applicable issues

- fixes #29

### Additional information

i'd be happy to do the other charts as well, i just figured i'd start with one and clear up any issues/questions first.

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.

### DIFF
--- a/charts/valkey/templates/serviceaccount.yaml
+++ b/charts/valkey/templates/serviceaccount.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "valkey.serviceAccountName" . }}
+  labels:
+      {{- include "valkey.labels" . | nindent 4 }}
+  {{- $annotations := merge .Values.serviceAccount.annotations .Values.commonAnnotations }}
+  {{- with $annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+automountServiceAccountToken: {{ .Values.serviceAccount.automountToken | default false }}
+{{- end }}
--- a/charts/valkey/values.yaml
+++ b/charts/valkey/values.yaml
@@ -92,6 +92,17 @@ service:
   ## @param service.annotations Service annotations
   annotations: {}
 
+## @section Service Account configuration
+## @param serviceAccount.create Enable creation of a Service Account
+serviceAccount:
+  create: true
+  ## @param serviceAccount.name Name of the Service Account
+  name: ""
+  ## @param serviceAccount.annotations Annotations to add to the Service Account
+  annotations: {}
+  ## @param serviceAccount.automountToken Enable automounting of the Service Account token
+  automountToken: false
+
 ## @section Ingress configuration
 ingress:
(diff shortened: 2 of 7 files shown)


charts/valkey/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "valkey.serviceAccountName" . }}
+  labels:
```
dhemeier, line 7: We need to add a namespace. In some scenarios like ArgoCD/FluxCD deployments, the `namespace` can be missing and so the ServiceAccount would be created in the `default` namespace.

Please add namespace without `quotes`:

```
metadata:
  name: {{ include "valkey.serviceAccountName" . }}
  namespace: {{ .Release.Namespace }}
```

dhemeier (changes requested):
Hello @matt-mai,

thank you very much for your contribution to the Helm Charts. We appreciate your effort. 

Please update the small issue regarding the missing namespace in the ServiceAccount, and then we can proceed with merging the pull request.

Thanks again!

charts/valkey/values.yaml
```diff
@@ -92,6 +92,17 @@ service:
   ## @param service.annotations Service annotations
   annotations: {}
 
+## @section Service Account configuration
+## @param serviceAccount.create Enable creation of a Service Account
+serviceAccount:
+  create: true
```
dhemeier: The default should be `false`

dhemeier:
Ahh @matt-mai ,

sorry I have overlooked the default `serviceAccount.create: true` and your comments around this topic. I would prefer to not create a serviceAccount as default. All pods that don´t need a specific configuration should have used the default ServiceAccount with `automountTokens: false`.

charts/valkey/values.yaml
matt-mai: done, I wasn't aware that the default SA had automount disabled, but I just checked, and you're right. 
In that case I agree that it doesn't make sense to create it by default.

charts/valkey/templates/serviceaccount.yaml
matt-mai: fixed

dhemeier (approved):

PR #48, opened 2025-09-09
[rabbitmq]: Fix erlang cookie copy / access rights

### LINKED ISSUE #31: [Bug] RabbitMQ Helm chart: .erlang.cookie not created
### Name and Version

0.0.3

When deploying RabbitMQ using the Helm chart, the pod starts successfully the first time because it uses the environment variable for the cookie instead of a .erlang.cookie file. However, on pod restart or reboot, the .erlang.cookie file is not created or persisted at /var/lib/rabbitmq/.erlang.cookie, causing RabbitMQ to crash.

_No response_

_No response_

What is the actual behavior?

After restarting the RabbitMQ pod, it fails to start because the .erlang.cookie file is missing from /var/lib/rabbitmq/.erlang.cookie.
As a result, RabbitMQ crashes with an error similar to:

`Cookie file /var/lib/rabbitmq/.erlang.cookie must be accessible by owner only`

### Proposed fix 
A quick fix could be to add an initContainer that explicitly creates the .erlang.cookie file if it doesn't already exist, by copying it from a mounted secret volume, and then setting the correct ownership and permissions. This ensures the cookie file is present before RabbitMQ starts.

Here are two snippets to add to the statefulset.yaml:

```
initContainers:
  - name: init-erlang-cookie
    image: busybox
    command:
      - sh
      - -c
      - |
        if [ ! -f /var/lib/rabbitmq/.erlang.cookie ]; then
          cp /cookie/erlang.cookie /var/lib/rabbitmq/.erlang.cookie && \
          chmod 400 /var/lib/rabbitmq/.erlang.cookie && \
          chown 999:999 /var/lib/rabbitmq/.erlang.cookie
        fi
    volumeMounts:
      - name: data
        mountPath: /var/lib/rabbitmq
      - name: erlang-cookie
        mountPath: /cookie
```

and 

```
volumes:
  - name: erlang-cookie
    secret:
      secretName: {{ include "rabbitmq.secretName" . }}
      items:
        - key: {{ include "rabbitmq.secretErlangCookieKey" . }}
          path: erlang.cookie
```

and increase the initialDelaySeconds of the readinessProbe to 10 seconds in values.yaml file

Thx

zOnlyKroks:
### Description of the change

Always copy the erlang cookie, giving it the right permissions

### Benefits

Running RabbitMQ after restart

### Applicable issues

- fixes #31 

### Checklist

### DIFF
--- a/charts/rabbitmq/templates/statefulset.yaml
+++ b/charts/rabbitmq/templates/statefulset.yaml
@@ -39,6 +39,7 @@ spec:
             - |
               echo "$RABBITMQ_ERLANG_COOKIE" > /var/lib/rabbitmq/.erlang.cookie
               chmod 400 /var/lib/rabbitmq/.erlang.cookie
+              chown {{ .Values.securityContext.runAsUser | default 999 }}:{{ .Values.securityContext.runAsGroup | default 999 }} /var/lib/rabbitmq/.erlang.cookie
           env:
             - name: RABBITMQ_ERLANG_COOKIE
               valueFrom:
--- a/charts/rabbitmq/templates/statefulset.yaml
+++ b/charts/rabbitmq/templates/statefulset.yaml
@@ -39,7 +39,6 @@ spec:
             - |
               echo "$RABBITMQ_ERLANG_COOKIE" > /var/lib/rabbitmq/.erlang.cookie
               chmod 400 /var/lib/rabbitmq/.erlang.cookie
-              chown 999:999 /var/lib/rabbitmq/.erlang.cookie
           env:
             - name: RABBITMQ_ERLANG_COOKIE
               valueFrom:
--- a/charts/rabbitmq/templates/statefulset.yaml
+++ b/charts/rabbitmq/templates/statefulset.yaml
@@ -49,7 +49,7 @@ spec:
           volumeMounts:
             - name: data
               mountPath: /var/lib/rabbitmq
-      {{- end }}‚
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           {{- if .Values.securityContext }}
--- a/charts/rabbitmq/templates/statefulset.yaml
+++ b/charts/rabbitmq/templates/statefulset.yaml
@@ -25,8 +25,10 @@ spec:
       {{ . | nindent 6 }}
       {{- end }}
       serviceAccountName: {{ include "rabbitmq.serviceAccountName" . }}
+      {{- with .Values.podSecurityContext }}
       securityContext:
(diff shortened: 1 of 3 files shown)


charts/rabbitmq/templates/statefulset.yaml
```diff
@@ -21,40 +21,41 @@ spec:
...
-        - name: {{ .Chart.Name }}-init
+        - name: init-erlang-cookie
           image: busybox
+          command:
+            - sh
+            - -c
+            - |
+              echo "$RABBITMQ_ERLANG_COOKIE" > /var/lib/rabbitmq/.erlang.cookie
+              chmod 400 /var/lib/rabbitmq/.erlang.cookie
+              chown 999:999 /var/lib/rabbitmq/.erlang.cookie
```
fwa-wup: should be user-id configured in securityContext: https://github.com/CloudPirates-io/helm-charts/blob/bb8e730cc6e5651412d8c1d7e2c23b96425b961c/charts/rabbitmq/values.yaml#L189

charts/rabbitmq/templates/statefulset.yaml
```diff
@@ -21,40 +21,41 @@ spec:
...
-                echo $RABBITMQ_ERLANG_COOKIE > /var/lib/rabbitmq/.erlang.cookie
-              else
-                echo "Erlang cookie already exists."
-              fi
-              chmod 600 /var/lib/rabbitmq/.erlang.cookie
           volumeMounts:
             - name: data
               mountPath: /var/lib/rabbitmq
-      {{- end }}
+      {{- end }}‚
```
fwa-wup: typo?

zOnlyKroks: yes indeed

zOnlyKroks:
@fwa-wup you may review again

charts/rabbitmq/templates/statefulset.yaml
```diff
@@ -39,7 +39,6 @@ spec:
             - |
               echo "$RABBITMQ_ERLANG_COOKIE" > /var/lib/rabbitmq/.erlang.cookie
               chmod 400 /var/lib/rabbitmq/.erlang.cookie
-              chown 999:999 /var/lib/rabbitmq/.erlang.cookie
```
fwa-wup: use something like
```
chown {{ .Values.securityContext.runAsUser | default 999 }}:{{ .Values.securityContext.runAsGroup | default 999 }} /var/lib/rabbitmq/.erlang.cookie
```

fwa-wup (approved):

PR #103, opened 2025-09-16
[postgres] add support for extra env vars from secret

frami:
### Description of the change

Support for adding extra environment variables from an existing secret to the Postgres container.

### Benefits

For example, it can be used for providing credentials in an init script. We use it to create different users and schemas
while in the init phase.

### Applicable issues

- fixes #

### Checklist


zOnlyKroks:
Thank you @frami for the commit, i will merge this now :D

PR #121, opened 2025-09-18
Add label selection to issue templates

Wurstkatze:
### Description of the change

This adds a dropdown to the bug report and feature request issue templates for selecting Helm chart labels.

### Benefits

We don't have to add labels manually and don't miss any.
Improves visibility of issues.

### Possible drawbacks

Users might apply the wrong labels, in wich case we would need to manually change them like before.

### Additional information

If a new Helm chart is added, the new label needs to be manually added to the issue templates.


Wurstkatze:
@dloewen2 Squash this when merging, lots of test commits (:

PR #138, opened 2025-09-23
[redis] Add metrics section to the README

bl4ko:
### Description of the change

Updating readme of the redis chart with the `metrics` section.

### Benefits

[Artifacthub](https://artifacthub.io/packages/helm/cloudpirates-redis/redis) will also display metrics configuration section.


dloewen2:
Hi @bl4ko,
thanks for the documentation, can you add bump the version one patch-level, so that it will be correctly deployed? Then i would merge this.

Issue #147, opened 2025-09-23
Add existingClaim to redis chart


roxannelandry:
### Name and Version

Redis

### What is the problem this feature will solve?

Currently, the chart does not provide a way to use pre-created storage resources. For users who already have existing PersistentVolumes (PVs) or need to comply with specific storage policies in their infrastructure, the lack of support for pre-created storage limits flexibility and makes migration from other charts more difficult.

### Describe the solution you'd like?

I’d like the ability to specify and attach a pre-created PersistentVolumeClaim (PVC) to the chart’s deployment. Something similar to what is done in the Bitnami chart (example: [redis application.yaml snippet](https://github.com/bitnami/charts/blob/f25c8f4fbabdf66fa9094311cb94304a16fb698d/bitnami/redis/templates/master/application.yaml#L523)) would be ideal. This would allow us to simply reference existing storage without relying on dynamic provisioning.

**This would allow us to:**

- Reference existing storage without relying on dynamic provisioning.

- Reuse infrastructure that is already in place.

- Continue leveraging our existing snapshot process for those disks, which is essential for our backup and recovery strategy.

**Additional context**

We are currently using the Bitnami Redis chart with pre-created storage, but we want to migrate to the official redis image and your chart.

Supporting pre-created storage would make the migration path smoother and enable users to reuse existing infrastructure instead of provisioning new resources.

### What alternatives have you considered?

Forking the repo and adding support for pre-created storage ourselves. However, we believe this is a generally useful feature for the community.

### Affected Helm charts

redis

zOnlyKroks:
Good Morning @roxannelandry ,
I will look into this soon, you may also contribute it yourself if you want.

roxannelandry:
Alright i will do it in a fork and when this work i will send a PR. I also saw that they is no way to add sidecar will check for that also.

PR #149, opened 2025-09-23
[Redis] Add support for existing claim

roxannelandry:
### Description of the change

This change adds support for specifying and attaching a pre-created PersistentVolumeClaim (PVC) to the chart’s deployment.  
It introduces a new optional `existingClaim` parameter in `values.yaml`, allowing users to reuse their own PVCs instead of always relying on dynamically provisioned volumes.  

### Benefits

- Enables migration from other charts (e.g., Bitnami) that already support pre-created storage.  
- Allows users to reuse existing infrastructure and PVs, reducing the need to provision new disks.  
- Supports use cases where storage snapshots or backup/restore processes are already in place.  
- Provides more flexibility for organizations with strict storage policies.  

### Possible drawbacks

- If misconfigured, users may point to PVCs that do not meet the storage requirements for the application.  
- Adds a small amount of complexity to the `values.yaml` by introducing another storage option.  

### Applicable issues

- fixes [#147  ](https://github.com/CloudPirates-io/helm-charts/issues/147)

### Additional information

We rely on pre-created storage with an existing snapshot/backup process for our disks. This change makes it possible to migrate from Bitnami Redis charts to the official chart without losing this important functionality.  

### Checklist

- [X] Variables are documented in the `values.yaml` and added to the `README.md`.  
- [X] Title of the pull request follows this pattern `[<name_of_the_chart>] Add support for existing PVC`.

### DIFF
--- a/charts/redis/test-production-values.yaml
+++ b/charts/redis/test-production-values.yaml
@@ -34,7 +34,6 @@ persistence:
   enabled: true
   size: 20Gi
   storageClass: ""
-  existingClaim: ""
 
 # Production resource requirements for Redis instances
 resources:


charts/redis/test-production-values.yaml
```diff
@@ -34,6 +34,7 @@ persistence:
   enabled: true
```
zOnlyKroks, line 34: could you remove this file?

zOnlyKroks:
other than that, it looks good

roxannelandry:
Question what will happen if we are in replica mode? 
Just to make sure...

roxannelandry:
humm that's what i tough : 
Multi-Attach error for volume "pvc-f60f54eb-70f7-4450-a532-ad0968e2d5d3" Volume is already used by pod(s) my-redis-patate-0

roxannelandry:
For it to work we would need to change a lot of thing and split the chart in master/replicas to be able to say replicas persistence enable false.

zOnlyKroks:
> For it to work we would need to change a lot of thing and split the chart in master/replicas to be able to say replicas persistence enable false.

I would not mind the split personally, but also lets hear @dloewen2 opinion on it

dloewen2:
The added changes in the standalone architecture look fine, but as you mentioned the replication architecture is at the moment not completely worked out. 
I think the split to master/replicas would be one of the best options, but also one with the most work. :D

PR #190, opened 2025-09-30
[nginx] Fix/nginx used ingress backend port

pascalfreundsrf:
### Description of the change
Currently, the input resource uses the `.Values.service.port` field for the backend.
The values.yaml field must currently look as follows in order to function properly.

```yaml
  service:
    port: 80
    ports:
      - port: 80
        targetPort: http
        name: http
```

#### Current ingress.yaml
```yaml
{{- $svcPort := .Values.service.port -}}
kind: Ingress
spec:
  rules:
    - host: {{ .host | quote }}
      http:
        paths:
          - path:
            backend:
              service:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
```

#### New
This PR would use the service's first array as the backend port in the ingress.
`{{- $svcPort := (index .Values.service.ports 0).port }}`

This PR is just one idea for how the Ingress resource can be populated.
You may have something else in mind..?
CC @mms-gianni

For your information: Bitnami does something similar (they use the service port with the service name “http”).
https://github.com/bitnami/charts/blob/main/bitnami/nginx/templates/ingress.yaml#L31

If this change is approved, I would appreciate it if it were also implemented in the other helm charts for consistency.

### Benefits

- Cleaner values.yaml files

### Possible drawbacks

- Removal of the missleading and undocumented `.Values.service.port` field

### Applicable issues

- fixes #

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.

### DIFF
--- a/charts/nginx/Chart.yaml
+++ b/charts/nginx/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nginx
 description: Nginx is a high-performance HTTP server and reverse proxy.
 type: application
-version: 0.1.9
+version: 0.1.10
 appVersion: "1.29.1"
 keywords:
   - nginx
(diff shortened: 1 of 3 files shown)


dloewen2:
Hi @pascalfreundsrf, 
thanks for the fix, release should be available in a few minutes.

PR #216, opened 2025-10-03
[keycloak] Allow keycloak to have a relative path

rob-64:
### Description of the change

Allow keycloak to have a relative path

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.1.10
+version: 0.1.11
 appVersion: "26.3.4"
 keywords:
   - keycloak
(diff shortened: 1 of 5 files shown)


zOnlyKroks (approved):

jzaehrin:
I have an issue on my side, the default value `/` looks not working. Argocd generate an empty variable without value: 
```
         env:                                                                                                                                                                                                   
         - name: KC_HTTP_RELATIVE_PATH                                                                                                                                                                          
         - name: KC_BOOTSTRAP_ADMIN_USERNAME                                                                                                                                                                    
           value: admin                                                                                                                                                                                       
```

rob-64:
@jzaehrin do we need to default it to / instead of "" then ?

jzaehrin:
I don’t know, I set the field to "/" as workaround. I saw that the default is set to / but it looks not working, the quote function block the default case ? This variable is mandatory ?

PR #221, opened 2025-10-06
[keycloak] default http relative path to '/' to fix argocd deployment

rob-64:
### Description of the change

default http relative path to '/' to fix argocd deployment

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.2.0
+version: 0.2.1
 appVersion: "26.3.4"
 keywords:
   - keycloak
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -109,7 +109,7 @@ keycloak:
   ## @param keycloak.production Enable production mode
   production: false
   ## @param keycloak.httpRelativePath Set relative path for serving resources; must start with a /
-  httpRelativePath: ""
+  httpRelativePath: /
 
 ## @section Database Configuration
 database:


rob-64:
@jzaehrin can you test this?

jzaehrin:
This correctly set the `KC_HTTP_RELATIVE_PATH` to `/` if keycloak.httpRelativePath is not set. But this variables is mandatory for Keycloak ? If not, I think it is better to not define it.

dloewen2:
according to the keycloak documentation the default for the http-relative.path is set to `/`, so i think it would be fine to set it as a default in the helm-chart as well, correct? 
(See: https://www.keycloak.org/server/all-config#category-http)

rob-64:
yea, its already the default, but another user said he was getting errors using argoCD, so i thought it would be best to just default to / instead of ""

zOnlyKroks:
> yea, its already the default, but another user said he was getting errors using argoCD, so i thought it would be best to just default to / instead of ""

could you try out to reach him, if this actually fixes his bug? Otherwise i´d be fine to merge this

jjssn:
I was just on my way to open a PR. Got the error below before changing `keycloak.httpRelativePath: ""` to `keycloak.httpRelativePath: "/"`. 
```
keycloak Failed to run 'build' command. Configuration validation failed quarkus.management.root-path is defined as the empty String ("") which the following Converter considered to be null: io.smallrye.config.Converters$BuiltInConverter
``` 

As stated in Keycloak documentation https://www.keycloak.org/server/all-config?q=relative the **http-relative-path** default value is "/". 

I see people mentioned ArgoCD, I can add that I used the command:
```
helm upgrade keycloak oci://registry-1.docker.io/cloudpirates/keycloak \
--version 0.2.0 -n keycloak -f values.yaml
``` 
Can post my values.yaml if needed, I switched from another chart to this one the only special things is that I have is keycloak.proxyHeaders=xforwarded, using an existing external postgres and configured ingress. Uninstalled the old chart before switching to this one, however it's the same database.

zOnlyKroks:
> I was just on my way to open a PR. Got the error below before changing `keycloak.httpRelativePath: ""` to `keycloak.httpRelativePath: "/"`.
> 
> ```
> keycloak Failed to run 'build' command. Configuration validation failed quarkus.management.root-path is defined as the empty String ("") which the following Converter considered to be null: io.smallrye.config.Converters$BuiltInConverter
> ```
> 
> As stated in Keycloak documentation https://www.keycloak.org/server/all-config?q=relative the **http-relative-path** default value is "/".

oh well then, consider this fixed, i will merge this now

PR #373, opened 2025-10-13
[keycloak]: Implement TLS in keycloak

### LINKED ISSUE #340: [keycloak] add support for TLS certificates
### Name and Version

keycloak

Allow users to import their own TLS certificates (pem, ...) for keycloak

- add `tls.enabled`, `tls.truststorePath`, `tls.existingSecret`
- follow [Keycloak documentation](https://www.keycloak.org/server/enabletls)

_No response_

### Additional informations

_No response_

### Affected Helm charts

keycloak

zOnlyKroks:
### Description of the change

Implement keycloak with TLS certificates, certificates may be issued by a installed cert manager

### Benefits

Working TLS

### Possible drawbacks

Harder to setup, we do not install a cert manager for the user

### Applicable issues

- fixes #340

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.6.0
+version: 0.6.1
 appVersion: "26.3.4"
 keywords:
   - keycloak


dloewen2:
@zOnlyKroks, 
if i see this correct you only give the option to use an existing TLS certificate.
Wouldn't it make sense to also give the option to auto-create the certificate via a kubernetes secret (https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_tls/)

Not 100 % sure if it would work this way.

zOnlyKroks:
> @zOnlyKroks, if i see this correct you only give the option to use an existing TLS certificate. Wouldn't it make sense to also give the option to auto-create the certificate via a kubernetes secret (https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_tls/)
> 
> Not 100 % sure if it would work this way.

I implemented it, by optionally creating a tls certificate through the cert manager, if installed

PR #390, opened 2025-10-14
[keycloak] Add support for templating hostname and hostnameAdmin

gkubes:
### Description of the change

This change allows Helm Chart consumers to use templates within values.yaml so the KeyCloak `hostname` and `hostname-admin` can be injected

### Benefits

Consumers can add this KeyCloak chart as a dependency with their own application and have a single value to configure the hostname.

### Possible drawbacks

Unknown

### Applicable issues

N/A

### Additional information

N/A

### Checklist

### DIFF
--- a/charts/rabbitmq/CHANGELOG.md
+++ b/charts/rabbitmq/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog
 
 
+## 0.5.1 (2025-10-15)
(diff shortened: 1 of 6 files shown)


zOnlyKroks:
thank you for your contribution :)

zOnlyKroks (approved):

PR #393, opened 2025-10-15
[universal] Rework all schema json

zOnlyKroks:
### Description of the change

Automatically generate ALL schema jsons

### Benefits

Correct, properly formatted, schema jsons

### Possible drawbacks

No descriptions etc

### Applicable issues

- fixes #

### Checklist

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is _not necessary_ when the changes only affect README.md files.
- [ ] Variables are documented in the values.yaml and added to the `README.md`

### DIFF
--- a/charts/mariadb/CHANGELOG.md
+++ b/charts/mariadb/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## 0.5.1 (2025-10-15)
+
+* Bugfix/redis non cloudpirates prefix (#394) ([02e03ed](https://github.com/CloudPirates-io/helm-charts/commit/02e03ed))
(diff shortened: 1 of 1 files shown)


zOnlyKroks:
Now those look like proper schema jsons, finally

dloewen2:
While yes the values-schema are fully configured and the types are defined correctly, the types are not so detailed like it would be, when done "by hand". Like ingress.tls is now only an array and not defined with the correct structure the ingress tls needs to have to be correct. 

But this takes a lot of work from us and automates it and also it should be more stable this way than before, because a lot of times things would be forgotten. 
So in my opinion this is the better way to go.

dloewen2 (approved):

PR #402, opened 2025-10-16
[keycloak] realm import

tanguyfalconnet:
### Description of the change

Keycloak can create realm from json config file. Currently we can only define that json config file as string in keycloak chart. If we want to import from a dedicated volume (using extraVolume & extraVolumeMount), we must be able to enable importing without using local config.

Currently, when we enable import, it automatically use configFile var, and forbid us to use extraVolume (we can't mount 2 volumes on the same dir: `/opt/keycloak/data/import`)

This PR aims to mount configFile only if it is defined.

### Benefits

Better flexibility.

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/charts/rabbitmq/CHANGELOG.md
+++ b/charts/rabbitmq/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 
 
+## 0.5.3 (2025-10-17)
+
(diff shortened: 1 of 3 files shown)


mms-gianni:
@tanguyfalconnet i can see your point. Makes sense to me. 

Please bump the chart.yaml version.

tanguyfalconnet:
@mms-gianni done :)

PR #477, opened 2025-10-27
[universal] unify extraEnvVars in all charts

### LINKED ISSUE #430: Streamline environment variable handling
### Name and Version

universal

Hi all,

the environment handling differs between the charts. Some support a dict of `extraEnv` others have a list of `extraEnvVars`. `extraEnvVars` has also the advantage of supporting other values than `value` like `valueFrom`.

If it is ok I would like to implement the bitnami like `extraEnvVars` support for all charts or are there any reasons for the inconsistency?

_No response_

### Additional informations

_No response_

### Affected Helm charts

universal

dloewen2:
### Description of the change

This PR updates all charts to use extraEnvVars instead of extraEnv or other implementations that are currently implemented.

IMPORTANT: 
This update will probably be a breaking change for a lot of users that have custom env vars set. 

### Benefits

Configurations for extraEnv Variables are equal in all charts.

### Possible drawbacks

These changes can break current deployments of the charts if the values are not being adjusted to the new format.

### Applicable issues

- fixes #430 

### Checklist


mms-gianni:
Looks good to me. 

I think this possible breaking changes are worth it.

dloewen2:
Yeah i was thinking the same. After 1.0.0 we have to handle breaking changes differently.

PR #505, opened 2025-10-30
[universal] Test auto-label change

Wurstkatze:
This should not trigger the auto-label pipeline, since it's from a fork and won't work anyway.


Wurstkatze:
Nice, works (:

PR #592, opened 2025-11-12
[memcached] support change to statefulset

### LINKED ISSUE #581: Memcached: add option to use statefulset for stable hostnames
### Name and Version

memcached Chart version: 0.5.3 / appVersion: "1.6.39"

Allow to have stable hostnames for the memcached pods, which can be important for software that requires hard configured hostnames (e.g. Apache Druid: https://druid.apache.org/docs/latest/configuration/#memcached).

I'd imagine that adding the option to specify the kind of the resource created (Deployment vs. StatefulSet) would be a possible solution.

-

### Additional informations

-

### Affected Helm charts

memcached

dloewen2:
### Description of the change

This PR adds the support to change the deploymenttype to a StatefulSet. 
Also some templates had the namespace definition missing. 

### Benefits

Option to deploy memcached as a StatefulSet. All Resources have the namespace defined and will be deployed in this. 

### Applicable issues

- fixes #581

### Checklist

### DIFF
--- a/charts/memcached/Chart.yaml
+++ b/charts/memcached/Chart.yaml
@@ -39,7 +39,4 @@ annotations:
       url: https://www.cloudpirates.io
   artifacthub.io/changes: |2
         - kind: changed
-          description: "Add support for pod labels (#550)"
-          links:
-            - name: "Commit 6071930"
-              url: "https://github.com/CloudPirates-io/helm-charts/commit/6071930"
+          description: "Chart updated"
--- a/charts/etcd/CHANGELOG.md
+++ b/charts/etcd/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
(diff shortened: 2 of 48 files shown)


charts/memcached/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: memcached
 description: Memcached is an in-memory key-value store for small chunks of arbitrary data (strings, objects) from results of database calls, API calls, or page rendering.
 type: application
-version: 0.5.3
+version: 0.5.4
```
applike-ss: Technically this should be a minor version increase instead of a patch one, because it adds a feature. Am I wrong?

dloewen2: You are absolutely right.

zOnlyKroks (approved):

PR #615, opened 2025-11-17
[Redis] sentinel use hostnames

### LINKED ISSUE #595: Redis HA w/ Sentinel loses track of master with frequent restarts
### Name and Version

redis v0.13.3

I have notice this behavior running on EKS with Karpenter. Here is a way to simulate it:

1. Create a Redis cluster with `architecture: replication` and `sentinel.enabled: true`
2. Query for the Redis master with `redis-cli -h $sentinel -p 26379 SENTINEL get-master-addr-by-name mymaster`
3. Delete the master pod to force a cluster rebalance
4. Repeat steps 2 & 3 ad nauseum

A bit too long to post, but I use the `extraObjects` field to create an extra Deployment and Service to always point at the Redis primary node that accepts writes.

Sentinel responds with a valid entry across restarts.

Sentinel responds with the IP address for the current master, sometimes for a pod that no longer even exists.

### Additional informations

Official Redis docs recommend using DNS entries, and this would greatly help when running in a Karpenter-enabled Kubernetes environment where pod rotations can be common.

https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/#ip-addresses-and-dns-names

I believe the configuration here could be adjusted to bring it in-line with official guidance:

- https://github.com/CloudPirates-io/helm-charts/blob/967558f3a7d700510e310ce43d795833a93cb98f/charts/redis/templates/statefulset.yaml#L138-L141
- https://github.com/CloudPirates-io/helm-charts/blob/967558f3a7d700510e310ce43d795833a93cb98f/charts/redis/templates/statefulset.yaml#L444

### Affected Helm charts

redis

dloewen2:
### Description of the change

Use the pod hostname in the sentinel configuration instead of the ip-address

### Applicable issues

- fixes #595 

### Checklist

### DIFF
--- a/charts/ghost/CHANGELOG.md
+++ b/charts/ghost/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
+## 0.5.0 (2025-11-17)
+
+* Update charts/ghost/values.yaml ghost to v6.8.0 (minor) (#605) ([7ab26d7](https://github.com/CloudPirates-io/helm-charts/commit/7ab26d7))
+* chore: update CHANGELOG.md for merged changes ([96c472e](https://github.com/CloudPirates-io/helm-charts/commit/96c472e))
+* chore: update CHANGELOG.md for merged changes ([9923048](https://github.com/CloudPirates-io/helm-charts/commit/9923048))
+
 ## 0.4.6 (2025-11-13)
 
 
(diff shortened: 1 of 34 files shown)


Overflow0xFFFF:
I've tested this change within my local cluster with 3 replicas. Sentinel does a much better job of keeping track of masters across multiple pod rotations with these changes. I confirmed this by deleting 1 to 2 pods each time for about 10 iterations and confirming that a master is properly elected.

zOnlyKroks (approved):

dloewen2:
Perfekt. This was also my result from testing in a local kind-cluster.

PR #695, opened 2025-12-05
[redis] Fix Redis issue with immutableFields cause by the label addition on volumeClaimTemplate

varet80:
The volumeClaimTemplate in STS is an immutable field. The new labels causes issues with ugprades! which is forcing us to reinstall redis on every Chart upgrade.
I am changing the labels to podLabels, used for the container templates.

image

### DIFF
--- a/charts/redis/Chart.yaml
+++ b/charts/redis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: redis
 description: An open source, in-memory data structure store used as a database, cache, and message broker.
 type: application
-version: 0.16.4
+version: 0.16.5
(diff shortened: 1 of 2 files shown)


dloewen2:
Hi @varet80, 
thanks for the fix. 
Yes the change with the labels was not very clean, sorry for the inconvenience.

varet80:
@dloewen2 I noticed the Release was canceled (Action). Can we run this release?
As I suspect there was a confusion with the Readme.

We would like to rebuild our Redis'es with the fix.

dloewen2:
@varet80 Release was just triggered again. Should be online in a few seconds

PR #701, opened 2025-12-05
[redis] return fqdn for sentinel replicas lookup (#700)

raouf-haddada:
* [redis] use fqdn for sentinel redis replicas, to avoid problems when calling it cross namespaces
recognized replicas fqdn
```
$ redis-cli -p 26379 -h redis-cloud-pirates-headless.redis
redis-cloud-pirates-headless.redis:26379> sentinel replicas redis-master
1)  1) "name"
    2) "redis-cloud-pirates-2.redis-cloud-pirates-headless.redis.svc.cluster.local:6379"
    3) "ip"
    4) "redis-cloud-pirates-2.redis-cloud-pirates-headless.redis.svc.cluster.local"
    5) "port"
    6) "6379"
    7) "runid"
    8) "f352db09cc6a2bae2f13a9d6c397bfe47f9d1db9"
    9) "flags"
   10) "slave"
   11) "link-pending-commands"
   12) "0"
   13) "link-refcount"
   14) "1"
   15) "last-ping-sent"
   16) "0"
   17) "last-ok-ping-reply"
   18) "640"
   19) "last-ping-reply"
   20) "640"
   21) "down-after-milliseconds"
   22) "5000"
   23) "info-refresh"
   24) "6860"
   25) "role-reported"
   26) "slave"
   27) "role-reported-time"
   28) "14213"
   29) "master-link-down-time"
   30) "0"
   31) "master-link-status"
   32) "ok"
   33) "master-host"
   34) "redis-cloud-pirates-0.redis-cloud-pirates-headless.redis.svc.cluster.local"
   35) "master-port"
   36) "6379"
   37) "slave-priority"
   38) "100"
   39) "slave-repl-offset"
   40) "12411"
   41) "replica-announced"
   42) "1"
2)  1) "name"
    2) "redis-cloud-pirates-1.redis-cloud-pirates-headless.redis.svc.cluster.local:6379"
    3) "ip"
    4) "redis-cloud-pirates-1.redis-cloud-pirates-headless.redis.svc.cluster.local"
    5) "port"
    6) "6379"
    7) "runid"
    8) "aa47a57e2d5758bee67ee8bdb271cff3870e3341"
    9) "flags"
   10) "slave"
   11) "link-pending-commands"
   12) "0"
   13) "link-refcount"
   14) "1"
   15) "last-ping-sent"
   16) "0"
   17) "last-ok-ping-reply"
   18) "757"
   19) "last-ping-reply"
   20) "757"
   21) "down-after-milliseconds"
[...]

Issue #628, opened 2025-11-18
Redis: Sentinel announce-ip changed


jamie-teqplay:
### Name and Version

redis:0.14.4

### What steps will reproduce the bug?

Sentinel's announces ip has changed in the latest version it used to be the pod ip
and now its `pod`.`release-name`-headless

which works for all the applications in the cluster but not when trying to connect to it for local development

### What is the expected behavior?

that i can configure the announce ip 
so that i can set it to something that can be reached outside of my cluster

### What is the actual behavior?

its set to  `pod`.`release-name`-headless
that is not reachable from outside of the cluster

### Affected Helm charts

redis

dloewen2:
I created a PR to add the option to use the hostname or ip, like it was a few versions ago.

Jamie-de-Leest:
Yes this work for me

PR #715, opened 2025-12-09
[valkey] Port #615, #639, #701 & fix custom config never being used

clrxbl:
### Description of the change

Ports Redis Helm chart PRs #615, #639 and #701 to the Valkey chart. Also fixes a critical issue where the custom Valkey config is never being mounted so Helm chart config options are not being respected.

### Benefits

Sentinel failover is unreliable without these PRs and has impacted our production workloads as a result.

### Possible drawbacks

Default behaviour of announcing hostnames can potentially break something, but should be done going forward. Can be reverted by the user with Helm chart setting sentinel.config.announceHostnames=false. Refer to #628 for one user complaining about this behaviour change.

### Applicable issues

N/A but fixes Valkey Sentinel failover

### Additional information

N/A

### Checklist

### DIFF
--- a/charts/valkey/README.md
+++ b/charts/valkey/README.md
@@ -128,6 +128,7 @@ The following table lists the configurable parameters of the Valkey chart and th
 
 | Parameter                  | Description                                               | Default                   |
 | -------------------------- | --------------------------------------------------------- | ------------------------- |
+| `config.mountPath`         | Redis configuration mount path                            | `/etc/valkey`             |
 | `config.maxMemory`         | Maximum memory usage for Valkey (e.g., 256mb, 1gb)        | `""`                      |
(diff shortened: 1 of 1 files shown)


clrxbl:
One thing I was uncertain about is the usage of cloudpirates.namespace vs .Release.Namespace. In the Redis chart the cloudpirates variable seems to be used but here it seems to be Release.Namespace. I've kinda blindly find & replaced it with cloudpirates.namespace in this PR, but happy to revert depending on what the choice should be.

dloewen2 (commented):
Looks good to me, only thing missing for me is the value `config.mountPath` in the ReadMe.md

clrxbl:
> Looks good to me, only thing missing for me is the value `config.mountPath` in the ReadMe.md

resolved

dloewen2 (approved):

Issue #693, opened 2025-12-04
Optional customUser.secretKeys


mreiche:
### Name and Version

postgres:0.12.1

### What is the problem this feature will solve?

It is not possible to define the custom user secret key only like:

```yaml
  customUser:
    name: "my-user"
    database: "my-database"
    existingSecret: "my-external-secret"
    secretKeys:
        password: "psql-user-password"
```

which will result in:
```yaml
  - name: CUSTOM_DB
    valueFrom:
      secretKeyRef:
        name: my-external-secret
        key: ''
  - name: CUSTOM_USER
    valueFrom:
      secretKeyRef:
        name: my-external-secret
        key: ''
  - name: CUSTOM_PASSWORD
    valueFrom:
      secretKeyRef:
        name: my-external-secret
        key: psql-user-password
```

### Describe the solution you'd like?

Please make the secretKeys for `username` and `database` optional with fallback to the `customUser.name` and `customUser.database`.

That makes it a lot more easier to define username and database as convention (by overriding values) and store the actual database user secret in a secret only. Especially when your secret is defined by an ExternalSecretProvider.

### Affected Helm charts

postgres

dloewen2:
Hi @mreiche, If i understand you correctly, if i use the values you added as an example, at the end i got two secrets, one existingSecret `my-external-secret` and one new created Secret containing the `customUser.name` and `customUser.database` or am i misunderstanding?

mreiche:
@dloewen2 The secret reference keeps the same, but defining a secret reference means, that you need to put all custom user information in that secret like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-custom-user-secret
stringData:
  user: "my-user"
  database: "my-database"
  password: "secret"
```

You can reference it in the chart:
```yaml
customUser:
  existingSecret: "my-custom-user-secret"
  secretKeys:
    name: "user"
    database: "database"
    password: "password"
```

But if you have a Secret, that is dynamically generated by an ExternalSecretProvider and only contains the secret information:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-generated-external-secret
stringData:
  password: "secret"
```
You cannot mix it like:
```yaml
customUser:
  name: "my-user"
  database: "my-database"
  existingSecret: "my-generated-external-secret"
  secretKeys:
    password: "password"
```

dloewen2:
[...]

PR #719, opened 2025-12-10
[postgres] Add support for password secret reference only

### LINKED ISSUE #729: Postgres chart cannot create custom user with special characters
### Name and Version

postgres:main

Set custom user name like:
```yaml
customUser:
  name: "mein-user"
  database: "meine-datenbank"
```

Which results in
```
Creating custom user: mein-user and database: meine-datenbank                                                                                                                                                                                           
ERROR:  syntax error at or near "-" at character 17                                                                                                                                                                   
STATEMENT:  CREATE USER mein-user WITH PASSWORD 'katze';                                                                                                                                                               
```

Yes, see above

Quote and create the user even with special characters.

Crashloop on custom user creation.

### Additional informations

_No response_

### Affected Helm charts

postgres

mreiche:
### Description of the change

This change implements the feature request #693 by allowing to set `customUser.name` and `customUser.database` as static string fallbacks if `customUser.existingSecret` is defined but `customUser.secretKeys.name` or `.database` are not.

To enable that feature, you have to configure your custom user like:

```yaml
customUser:
  existingSecret: "my-password-only-secret"
  name: "my-username"
  database: "my-database"
  secretKeys:
    name: ""
    database: ""
    password: "my-password-key"
```

Tests are included in this PR.

### Benefits

This change enables the migration to existing secrets that only contain the password.

### Possible drawbacks

You have to enable `customUser.existingSecret` and disable unwanted `customUser.secretKeys`.

### Applicable issues

- implements #693
- fixes #729

### Additional information

What is the best practice to update the `values.schema.json` and `README.md` except manual?

### DIFF
--- a/charts/postgres/templates/statefulset.yaml
+++ b/charts/postgres/templates/statefulset.yaml
@@ -92,11 +92,11 @@ spec:
 {{ toYaml . | indent 12 }}
             {{- end }}
 
-            {{/* Custom user setup if a name or an existing secret is defined */}}
+            {{- /* Custom user setup if a name or an existing secret is defined */}}
             {{- if and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret) }}
             - name: CUSTOM_DB
 
-              {{/* Setup a secret ref if a key is defined */}}
+              {{- /* Setup a secret ref if a key is defined */}}
               {{- if .Values.customUser.secretKeys.database }}
               valueFrom:
                 secretKeyRef:
@@ -109,11 +109,9 @@ spec:
                   {{- end }}
                   key: {{ tpl .Values.customUser.secretKeys.database . | quote }}
               {{- else }}
-
-              {{/* Fallback to static value otherwise */}}
+              {{- /* Fallback to static value otherwise */}}
               value: {{ tpl .Values.customUser.database . | quote }}
               {{- end }}
-
             - name: CUSTOM_USER
               {{- if .Values.customUser.secretKeys.name }}
               valueFrom:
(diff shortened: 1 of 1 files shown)


charts/postgres/templates/statefulset.yaml
```diff
@@ -91,33 +91,43 @@ spec:
             {{- with .Values.extraEnvVars }}
 {{ toYaml . | indent 12 }}
             {{- end }}
+
+            {{/* Custom user setup if a name or an existing secret is defined */}}
```
dloewen2: Please add a `-` at the beginning of all your comments. This removes the empty lines in the build template. (Makes it a bit prettier in debugging :) )

dloewen2:
The values.schema.json will be updated automatically via a workflow. 
The ReadMe you currently have to update manually, as most generation tools can not support the additional text around the helm-values tables.

charts/postgres/templates/statefulset.yaml
mreiche: I tried to add a "-" to the comments, but this results in
```
Error: UPGRADE FAILED: parse error at (postgres/templates/statefulset.yaml:111): "-"
```

I don't see any option to remove these white lines except by removing the comments right now. Any ideas?

dloewen2: If i update the comments like this it works for me: `{{- /* Setup a secret ref if a key is defined */}}`

if that does not help, i will just merge it without.

mreiche: Ah nevermind. My IDE was misleading. It works.

mreiche:
@dloewen2 Please keep it mind that this also includes #729

dloewen2:
Yes, your functionality additions look good to me. That was the only thing i got as a review for improvement. :)

PR #723, opened 2025-12-10
[redis]: allow changing revisionHistoryLimit

Farfaday:
### Description of the change
allow changing revisionHistoryLimit
We keep 10 as the default value to make the change transparent and avoid creating a breaking change.

### Benefits

Allow changing the value of revisionHistoryLimit

### Possible drawbacks
Not known yet.

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is _not necessary_ when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md`
- [X] Title of the pull request follows this pattern [<name_of_the_chart>] Descriptive title

PR #725, opened 2025-12-10
[redis, valkey,rabbitmq,zookeeper]: allow setting revisionHistoryLimit

Farfaday:
### Description of the change

Similar to https://github.com/CloudPirates-io/helm-charts/pull/723
- feat: Allow setting revisionHistoryLimit for valkey and rabbitmq
- fix: Remove "| default 10" in helm templates, as it blocks user the possibility to set revisionHistoryLimit to 0.

### Benefits
Allow changing the value of revisionHistoryLimit

### Possible drawbacks
None, non-breaking change.

### Checklist

### DIFF
--- a/charts/postgres/CHANGELOG.md
+++ b/charts/postgres/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
+## 0.13.2 (2025-12-09)
+
+* chore: update CHANGELOG.md for merged changes ([aa25c3a](https://github.com/CloudPirates-io/helm-charts/commit/aa25c3a))
+* Fix non crashloopbackoff on broken scripts ([35a5175](https://github.com/CloudPirates-io/helm-charts/commit/35a5175))
+* chore: update CHANGELOG.md for merged changes ([a598f4d](https://github.com/CloudPirates-io/helm-charts/commit/a598f4d))
+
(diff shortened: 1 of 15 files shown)


charts/valkey/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: valkey
 description: High performance in-memory data structure store, fork of Redis. Valkey is an open-source, high-performance key/value datastore that supports a variety of workloads such as caching, message queues, and can act as a primary database.
 type: application
-version: 0.12.0
+version: 0.13.0
```
dloewen2, line 5: For the other Charts you only bumped the chart by a patch version. So i would do it here as well

dloewen2 (commented):
Please bump the rabbitmq chart version as well

charts/valkey/Chart.yaml
Farfaday: Where I added it as a new "feature", I bumped minor, and where I fixed the bad | default 10 behaviour, I bumped a patch version :)

dloewen2: Ah i understand. Perfekt then :)

dloewen2 (approved):

PR #730, opened 2025-12-11
[rabbitmq] Add externalTrafficPolicy and allocateLoadBalancerNodePorts to service spec

iTiamo:
### Description of the change

This PR adds the [externalTrafficPolicy](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip) and [allocateLoadBalancerNodePorts](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation) to the service spec for the RabbitMQ Helm chart.

### Benefits

1. Setting the externalTrafficPolicy property to Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type Services. This also sets the healthCheckNodePort value automatically if the Service type is LoadBalancer. This integrates well with, for example, Azure Load Balancer allowing Health Probes to be automatically configured correctly.
2. Setting the allocateLoadBalancerNodePorts property to false prevents allocating NodePorts to the Service. This is useful when a cloud load balancer is used that supports routing traffic directly to pods and if your cluster has quota set on NodePorts.

This functionality is also included in the [Bitnami Helm chart for RabbitMQ](https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/templates/svc.yaml#L32).

### Possible drawbacks

None. The default values are kept as the default, this will only allow users to overwrite those defaults.

### Applicable issues

None.

### Additional information

None.

### Checklist

### DIFF
--- a/charts/postgres/CHANGELOG.md
+++ b/charts/postgres/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
+## 0.13.2 (2025-12-09)
+
+* chore: update CHANGELOG.md for merged changes ([aa25c3a](https://github.com/CloudPirates-io/helm-charts/commit/aa25c3a))
+* Fix non crashloopbackoff on broken scripts ([35a5175](https://github.com/CloudPirates-io/helm-charts/commit/35a5175))
+* chore: update CHANGELOG.md for merged changes ([a598f4d](https://github.com/CloudPirates-io/helm-charts/commit/a598f4d))
+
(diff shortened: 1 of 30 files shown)


dloewen2:
Hi @iTiamo, 
thanks for the PR. After you update the version i would merge this.

iTiamo:
Thx, LGTM

dloewen2 (approved):

PR #731, opened 2025-12-11
fix(redis): prevent password logging in sentinel startup

### LINKED ISSUE #728: Redis: sentinel.conf dumped to console includes password
### Name and Version

Redis 0.17.0

When using sentinel auth, the chart dumps the passwords, as part of its dump of the config.

requirepass "xxxxxxxxxxxxx" 
sentinel auth-pass mymaster "xxxxxxxxxxxxx"

Root cause is this line (which is also in the valkey chart)
https://github.com/CloudPirates-io/helm-charts/blob/0bbe33d58e9c34600300b25e6ae29d37dedb0b66/charts/redis/templates/statefulset.yaml#L497

auth:
  enabled: true
  sentinel: true
  existingSecret: vault-redis-password
  existingSecretPasswordKey: redis-password

The config is not always dumped. It would be ok on debug or info level.

secrets are dumped to the console and from there get collected to logging system.  

### Additional informations

_No response_

### Affected Helm charts

redis, valkey

cdm-arm:
## Summary
- Add `sentinel.loglevel` configuration option to control Sentinel log verbosity
- Only log sentinel config (which contains passwords) when loglevel is set to `debug`
- Default behavior (notice/warning) now prints only "Starting Sentinel..." without exposing config

## Problem
When Redis Sentinel is configured with authentication (`auth.enabled: true` and `auth.sentinel: true`), the startup script logs the entire `sentinel.conf` file including:
- `requirepass <password>`
- `sentinel auth-pass mymaster <password>`

This exposes Redis passwords in pod logs.

## Solution
Make config logging conditional on the new `sentinel.loglevel` setting:
- `debug`: Full config logged (for troubleshooting)
- `verbose`/`notice`/`warning`: Only "Starting Sentinel..." message (secure default)

## Changes
- `charts/redis/values.yaml`: Add `sentinel.loglevel` parameter (default: `notice`)
- `charts/redis/templates/statefulset.yaml`: Use configurable loglevel and conditional config logging

## Testing
```bash
# Default (notice) - should NOT log config
helm template redis ./charts/redis --set sentinel.enabled=true --set auth.enabled=true --set architecture=replication | grep "Starting Sentinel"
# Output: echo "Starting Sentinel..."

# Debug - should log config
helm template redis ./charts/redis --set sentinel.enabled=true --set auth.enabled=true --set architecture=replication --set sentinel.loglevel=debug | grep -A1 "Starting Sentinel"
# Output: echo "Starting Sentinel with config:"
#         cat /tmp/sentinel.conf
```

Fixes #728

### DIFF
--- a/charts/postgres/CHANGELOG.md
+++ b/charts/postgres/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
+## 0.13.2 (2025-12-09)
+
+* chore: update CHANGELOG.md for merged changes ([aa25c3a](https://github.com/CloudPirates-io/helm-charts/commit/aa25c3a))
+* Fix non crashloopbackoff on broken scripts ([35a5175](https://github.com/CloudPirates-io/helm-charts/commit/35a5175))
+* chore: update CHANGELOG.md for merged changes ([a598f4d](https://github.com/CloudPirates-io/helm-charts/commit/a598f4d))
+
 ## 0.13.1 (2025-12-09)
 
 * chore: update CHANGELOG.md for merged changes ([b841a05](https://github.com/CloudPirates-io/helm-charts/commit/b841a05))
--- a/charts/postgres/Chart.lock
+++ b/charts/postgres/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: oci://registry-1.docker.io/cloudpirates
-  version: 2.0.0
-digest: sha256:ae9378e0dcfd09a35b7f994007db99c2d6fe02ef7634f424d5233237c209a1c7
-generated: "2025-10-14T11:14:49.235251+02:00"
+  version: 2.1.0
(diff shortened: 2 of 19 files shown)


charts/redis/values.yaml
```diff
@@ -308,6 +308,9 @@ sentinel:
   quorum: 2
   ## @param sentinel.downAfterMilliseconds Time in milliseconds after the master is declared down
   downAfterMilliseconds: 1500
+  ## @param sentinel.loglevel Sentinel log level. Allowed values: debug, verbose, notice, warning
+  ## When set to 'debug', full sentinel config (including auth) will be logged at startup
+  loglevel: notice
```
dloewen2, line 313: Please add this new Values to the Readme.md. 
Currently we don't have this step automated.

dloewen2:
Also with your other PR merged, please bump the version in here.

charts/redis/values.yaml
cdm-arm: Done

PR #741, opened 2025-12-12
[minio] Rename post job and add variables to configure it

ixxeL2097:
### Description of the change

The minio `post-job` `Job` that I initially proposed and merged is not meaningful in terms of naming and would gain clarity by being renamed to reflect its actual action. Additionally, some variables have been added to make it more customizable and provide users with more granularity in their configuration.

On our side, we had scenarios where we needed to customize these parameters and it turned out to be helpful.

### Benefits

- Provide clarity regarding deployed objetcs.
- Provide more configuration options for infrastructure needs.
- Add dedicated section to later improve the related `Job` configuration.

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/charts/minio/Chart.yaml
+++ b/charts/minio/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: minio
 description: High Performance Object Storage compatible with Amazon S3 APIs
 type: application
-version: 0.5.7
+version: 0.6.0
 appVersion: "2025.09.07"
 keywords:
   - minio
--- a/charts/minio/templates/bucket-init-configmap.yaml
+++ b/charts/minio/templates/bucket-init-configmap.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "minio.fullname" . }}-post-job
+  name: {{ include "minio.fullname" . }}-bucket-init
   namespace: {{ .Release.Namespace }}
   labels:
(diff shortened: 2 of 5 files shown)


ixxeL2097:
I don't really understand why the Generate values.schema.json workflow fail

Any idea @zOnlyKroks ?

zOnlyKroks:
> I don't really understand why the Generate values.schema.json workflow fail
> 
> Any idea @zOnlyKroks ?

there's a bug currently in it, nothing wrong with your code :)

dloewen2:
@ixxeL2097 was there a reason that you did not add the helm-hook annotations or was this something overlooked?

ixxeL2097:
> @ixxeL2097 was there a reason that you did not add the helm-hook annotations or was this something overlooked?

For this specific PR, I don't remember, maybe I forgot to integrate. But generally speaking, we tend to avoid Helm hooks as they present several issues. First, they're somewhat against Kubernetes philosophy to orchestrate deployment order, and second, they don't always work well with some other technologies (ArgoCD, GCP Marketplace...). We tend to use `initContainers` or similar patterns whenever possible, with strong healthchecks, making it easier to apply manifests with Kubernetes native features rather than relying on Helm-specific ones.

So, whenever I contribute back to your repository guys, I will always try to stick with this rule.

EDIT: We are using Minio in some of our deployments and we are in the process of migrating from helm hooks to full Kubernetes native manifests (without any kind of tech specific hook or something like that).

PR #750, opened 2025-12-15
fix(redis): Always set masterauth for non-standalone architectures

cdm-arm:
## Summary
When a pod starts as master but later becomes a replica (e.g., after sentinel failover), it needs `masterauth` to authenticate to the new master during replication.

## Problem
Previously, `masterauth` was only added when a pod was initially configured as a replica. This caused NOAUTH errors when:
1. Pod starts as master (no masterauth in config)
2. Sentinel promotes another pod as master
3. Original master receives SLAVEOF command from sentinel
4. Replication fails: `MASTER aborted replication with an error: NOAUTH Authentication required.`

### Evidence from production logs
```
# redis-cli -a "$REDIS_PASSWORD" CONFIG GET masterauth
1) "masterauth"
2) ""   # Empty! This pod started as master so masterauth was never set
```

After manual fix:
```
# redis-cli -a "$REDIS_PASSWORD" CONFIG SET masterauth "$REDIS_PASSWORD"
OK
# Replication immediately recovers
```

## Solution
Always add `masterauth` to the Redis config for non-standalone architectures (`replication`, `cluster`), regardless of initial role. This ensures any pod can become a replica after failover.

The fix adds `masterauth` unconditionally after all role-specific configuration but before user-provided config, so:
- All pods (master or replica) have masterauth configured
- User config can still override if needed
- No impact on standalone architecture (no replication)

## Testing
- [x] `helm lint` passes
- [x] Template renders correctly with `sentinel.enabled=true` and `auth.enabled=true`
- [x] Verified masterauth appears in config for both initial masters and replicas

## Related Issues
This fixes the root cause of NOAUTH replication errors that occur after sentinel failover in authenticated Redis Sentinel deployments.

### DIFF
--- a/charts/redis/Chart.yaml
+++ b/charts/redis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: redis
 description: An open source, in-memory data structure store used as a database, cache, and message broker.
 type: application
-version: 0.17.3
+version: 0.17.4
 appVersion: "8.4.0"
 keywords:
   - redis


zOnlyKroks (approved):
thank you for your contribution

cdm-arm:
@zOnlyKroks - I just saw that 594cf3d286a9fb09f428eea864001cf9c4ca5b9b already bumped the version to 0.17.4 - So the merge seemed to not trigger a new release. Should I create a 0.17.5 update for this PR / make a new one?

PR #811, opened 2026-01-07
[redis] Include Sentinel port in default service when enabled

raouf-haddada:
### Description of the change

Include the sentinel port in the default redis service, it's missing when migrating from old bitnami helm charts.


zOnlyKroks:
Kindly bump the chart version again, and this can be merged :)

PR #829, opened 2026-01-12
[mongodb] add replication architecture

dloewen2:
### Description of the change

With these changes the replication architecture for mongodb can be used. 

**BREAKING CHANGE**: This changes some immutable fields of the StatefulSet, so the update to this version can not be done completely automated.

### Benefits

Adds replication architecture for users.

### Possible drawbacks

More points of failure with the newly added init-container, custom-scripts and configurations. 
To update to this version you will need to delete the current statefulSet, as there are changes to immutable fields. 

### Possible way to upgrade 

To Upgrade to this version and keep your data you can delete your currently deployed StatefulSet with the `--cascade=orphan` parameter. With this the currently running Pod will be kept alive with the PVC still Attached. After the deletion apply the current version of the StatefulSet and delete the old Pod. 
The new StatefulSet will create a pod and attach the existing PVC. 

### Checklist

### DIFF
--- a/charts/rabbitmq/CHANGELOG.md
+++ b/charts/rabbitmq/CHANGELOG.md
@@ -44,6 +44,14 @@
 
 * No changes recorded
 
+## 0.11.5 (2026-01-12)
+
+* queuebalancing schema types (#831) ([f1f1fbf](https://github.com/CloudPirates-io/helm-charts/commit/f1f1fbf))
+* chore: update CHANGELOG.md for merged changes ([e40975d](https://github.com/CloudPirates-io/helm-charts/commit/e40975d))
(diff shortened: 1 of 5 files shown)


i45Ralph4545:
@dloewen2 If you allow for extra volume & volume mounts in the init container, we can make use of the init-script properly:

if [ -d /extrascripts ]; then
        echo "Copy extra scripts"
        cp /extrascripts/* /initscripts
      fi
      if [ -d /customscripts ]; then
        echo "Copy custom scripts"
        cp /customscripts/* /initscripts
      fi

dloewen2:
@i45Ralph4545 
thanks for the input. I'm gonna implement that in the next few days. Should not too much work.

i45Ralph4545:
@dloewen2 great, thank you

PR #837, opened 2026-01-13
[postgres] Add support for docker hardened images

Loupeznik:
### Description of the change

Adds option to define postgres container startup parameters different from the default ones provided by the helm chart and adjusts storage paths for non-default docker images. This mainly aims to resolve compatibility problem with Docker Hardened Images (DHI)

### Benefits

Adds compatibility with Docker Hardened Images and other hardened postgres images where the startup of the postgres container is different than in the original library postgres docker image (see [docs](https://hub.docker.com/hardened-images/catalog/dhi/postgres/guides))

### Possible drawbacks

Unknown at this time

### Applicable issues

None

### Checklist

### DIFF
--- a/charts/postgres/Chart.yaml
+++ b/charts/postgres/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: postgres
 description: The World's Most Advanced Open Source Relational Database
 type: application
-version: 0.13.9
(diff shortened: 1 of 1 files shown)


zOnlyKroks:
I will check this later, but from a quick glance it looks good

PR #859, opened 2026-01-19
[keycloak]: Move to stateful set

### LINKED ISSUE #840: Keycloak: Use StatefulSet instead of Deployment
### Name and Version

keycloak 0.11.14

Install keycloak using the "High Availability Setup" example on https://artifacthub.io/packages/helm/cloudpirates-keycloak/keycloak

Add a `LoadBalancer` service:

```
apiVersion: v1
kind: Service
metadata:
  name: keycloak-external
  namespace: default
  labels:
    app.kubernetes.io/name: keycloak-external
spec:
  type: LoadBalancer
  ports:
    - port: 7081
      targetPort: 7081
      protocol: TCP
      name: http
  selector:
    app.kubernetes.io/name: keycloak
```

Check the contents of table `migration_model` in the keycloak database and open http://localhost:7081 in a browser.

_No response_

The keycloak login page renders correctly and the table `migration_model` contains only one row.

The login page doesn't render correctly and when refreshing the page there are many HTTP 404 errors in the console.

In table `migration_model` in the keycloak database you can see multiple rows with the same `version`.

### Additional informations

As stated in the following two links, keycloak in "High Availability Mode" with more than one replica should be deployed as `StatefulSet`, not as `Deployment` as it is currently done here:

- https://github.com/keycloak/keycloak/issues/39866#issuecomment-2911891340 ("Core Problem")
- https://www.keycloak.org/getting-started/getting-started-kube#_start_keycloak

In `Deployments`, the Pods are started in parallel which causes issues during database schema migration/initialization as keycloak doesn't lock the database during migration. 

When using `StatefulSet`, the Pods are started one after another and migration/initialization works as intended.

The issues doesn't appear, when using `kubectl port-forward` because this will route all requests to the same Pod. 

The `LoadBalancer` randomly routes the requests to different Pods and each Pod uses a different resource id from the `migration_model`. This leads to random 404 errors, depending on which Pod received the request. 

Setting `sessionAffinity: ClientIP` on the LoadBalancer would also mitigate the issue, but is not a real solution.

### Affected Helm charts

keycloak

zOnlyKroks:
### Description of the change

Migrate keycloak from deployment to statefulset

### Benefits

See #840 

### Applicable issues

- fixes #840

### Additional information

Its a breaking change

### Checklist

### DIFF
--- a/charts/keycloak/CHANGELOG.md
+++ b/charts/keycloak/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
+## 0.11.5 (2026-01-19)
+
+* [keycloak/keycloak] Update charts/keycloak/values.yaml keycloak/keycloak to v26.5.1 (patch) (#860) ([8b8713d](https://github.com/CloudPirates-io/helm-charts/commit/8b8713d))
+* chore: update CHANGELOG.md for merged changes ([2b47b4b](https://github.com/CloudPirates-io/helm-charts/commit/2b47b4b))
+* chore: update CHANGELOG.md for merged changes ([f764d87](https://github.com/CloudPirates-io/helm-charts/commit/f764d87))
+* chore: update CHANGELOG.md for merged changes ([de14f91](https://github.com/CloudPirates-io/helm-charts/commit/de14f91))
+
 ## 0.11.4 (2026-01-08)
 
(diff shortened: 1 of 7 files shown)


vonhutrong:
Hi,
Because "Its a breaking change", do you think should there be a migration note?

PR #868, opened 2026-01-19
Explicitly set database name when creating custom user

lucendio:
Creating a `customUser` fails with *'database "<auth.username>" does not exist'* if `auth.username` is (a) set and (b) to sth other than `'postgres'`.

It appears that if no database is defined for the `psql` command, it defaults to the value of `--username`. Explicitly setting the database solves the issue.


dloewen2:
Hi @lucendio, 
thanks for the contribution. Could you update the chart version to 0.14.3?
Then i would merge you PR.

PR #869, opened 2026-01-19
POSTGRES_DB is empty if auth.database is not overridden

lucendio:
Use the proper partial to set `POSTGRES_DB`. This change means (a) actually following the existing pattern that is already in use, and (b) to allow successfully overriding the default database.


dloewen2:
@lucendio 
Looks good to me. If you also update the chart-verison here i can merge it. 
So with your other PR, you can set this version to 0.14.4

PR #878, opened 2026-01-21
[postgres] Add support for volumeName

mreiche:
### Description of the change

This change allows to override the default volume name `data` to something else.
That enables the `StatefulSet` to override take over already existing PVCs.

### Benefits

If your previous `StatefulSet` created PVCs name like `pgdata-myservice-0`, you were unable to re-claim that by migrating to this chart, because the volume name was hardcoded to `data`, which results in PVC names like `data-myservice-0`.

This change makes it a lot easier to migrate from other charts and perform migration on the same PVC using initContainers.

### Possible drawbacks

- None

### Applicable issues

- None file

### Additional information

- None

### Checklist


mreiche:
It seems that the failing check comes from a temporary problem.

zOnlyKroks:
> It seems that the failing check comes from a temporary problem.

yes, we dont currently know why it fails, but its harmless, so no need to worry :)

PR #880, opened 2026-01-21
[mongodb]: enable replicaSet.[key,keySecretName]

lucendio:
As their respected default values were not set, it normally wouldn't fail, but if one of them is set accordingly (to string), helm complains when rendering charts like so:

```
  Error: values don't meet the specifications of the schema(s) in the following chart(s):
  mongodb:
  - at '/replicaSet/keySecretName': got string, want null
  - at '/replicaSet/key': got string, want null
```

Setting each type to string allows to set and override each value. Please note, that conditionals referring to those values will still be falsy by default as their default value is an empty string.

`key` may be a multi-line string. Using `quote` enables line-break activates.

### DIFF
--- a/charts/mongodb/Chart.yaml
+++ b/charts/mongodb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mongodb
 description: MongoDB a flexible NoSQL database for scalable, real-time data management
 type: application
-version: 0.9.1
+version: 0.9.2
 appVersion: "8.2.3"
 keywords:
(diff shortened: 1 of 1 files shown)


zOnlyKroks:
Would you kindly bump the Chart version, that would be the last thing missing here :) (Please do so with your other pr #881 in mind)

zOnlyKroks:
I am going to merge this, but keep in mind, as of today, we normally enforce that all commits are signed with a gpg/ssh key :)

PR #881, opened 2026-01-22
[mongodb]: introduce new value replicaSet.extraInit.disableTelemetry

lucendio:
By default, `mongosh` telemetry is enabled. This new value allows to disable it. For more details, please refer to the link in the parameter description.


zOnlyKroks:
Would you kindly bump the Chart version, that would be the last thing missing here :) (Please do so with your other pr #880 in mind)

zOnlyKroks:
I am going to merge this, but keep in mind, as of today, we normally enforce that all commits are signed with a gpg/ssh key :)

PR #894, opened 2026-01-23
[redis] Fix network policy egress, improve cluster init job reliability

i5okie:
This PR addresses several issues in the Redis chart:

1. **Network Policy Egress Fix:** When `egressEnabled: false`, now uses `egress: [{}]` (allow all) instead of removing `Egress` from policyTypes. This (actually) fixes compatibility with OpenShift SDN.

2. **Cluster Init Job Improvements**:
   - Added idempotency check - skips cluster creation if already initialized (safe for post-upgrade hook)
   - Added pod template metadata with labels for proper identification and NetworkPolicy matching
   - Fixed annotations indentation (`indent` → `nindent`)

3. **Consistency Fixes**:
   - Sentinel master-discovery deployment now uses `cloudpirates.render*SecurityContext` helpers
   - Added explicit NetworkPolicy rule for cluster-init job access

### Benefits
- OpenShift SDN compatibility for egress configuration
- Cluster init job no longer fails on helm upgrade when cluster already exists
- Job pods are now properly labeled and discoverable

### Possible drawbacks
None. Changes are backward compatible.

### DIFF
--- a/charts/ghost/Chart.yaml
+++ b/charts/ghost/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: ghost
 description: A simple, powerful publishing platform that allows you to share your stories with the world.
 type: application
-version: 0.8.0
-appVersion: "6.13.2"
+version: 0.8.1
+appVersion: "6.14.0"
 keywords:
   - ghost
   - blogging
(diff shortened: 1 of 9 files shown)


zOnlyKroks:
Mind bumping the Chart version for me? Otherwise looks good

PR #901, opened 2026-01-26
[rabbitmq] Add detailed metrics support (#804)

### LINKED ISSUE #804: RabbitMQ: Add support per object / detailed metrics
### Name and Version

rabbitmq 0.10.1

By default the rabbitmq_prometheus plugin comes with aggrageted and per-object metrics;

https://www.rabbitmq.com/docs/prometheus#metric-aggregation

> RabbitMQ can return Prometheus metrics in two modes:
> **Aggregated**: metrics are aggregated by name. This mode has lower performance overhead with the output size constant, even as the number of objects (e.g. connections and queues) grows.
> **Per-object**: individual metric for each object-metric pair. With a large number of stats-emitting entities, e.g. a lot of connections and queues, this can result in very large payloads and a lot of CPU resources spent serialising data to output.

Currently the rabbitmq chart only adds a service monitor for the aggregated metrics using the /metrics endpoint.  The downside of this there is no easy option to have the service monitor scrape the additional metrics endpoints

I would like to have an option to easily configure additional service monitor endpoints such as /metrics/per-object and /metrics/detailed  with additional family and vhost params. 

Maybe something like: 
```yaml
metrics:
  enabled: true
  serviceMonitor:
     detailed:
        enabled: true
        family: 
           - ..
           - ..
        vhost: 
           - ..
``` 

Or maybe an "extraServiceMonitorEndpoints" option somewhat similar like `extraEnvVars` or `extraVolumeMounts`

It is possible to create a second service monitor using the `extraObjects` but including different templates /includes becomes challeging due the fact all template references within the extra objects needs to be quoted. 

```yaml
  extraObjects:
    - apiVersion: monitoring.coreos.com/v1
      kind: ServiceMonitor
      metadata:
        name: rabbitmq-detailed
        namespace: "{{ .Release.Namespace }}"
        labels:
          release: kube-prometheus-stack
      spec:
        endpoints:
          - path: /metrics/detailed
            params:
              family:
                - queue_coarse_metrics
                - queue_consumer_count
            port: prometheus
            interval: "30s"
            scrapeTimeout: "10s"
        namespaceSelector:
          matchNames:
            - "{{ .Release.Namespace }}"
        selector:
          matchLabels:
            app.kubernetes.io/name: '{{- include "rabbitmq.name" . }}'
            app.kubernetes.io/instance: "{{ .Release.Name }}"
```

### Additional informations

https://www.rabbitmq.com/docs/prometheus#metric-aggregation

I think it is most preferred to actually use the aggregated- (`/metrics`)  and detailed (`/metrics/detailed`)  endpoint and not perse the per object (`/metrics/per-object`) endpoint, as latter comes with a performance impact on large RabbitMQ setups

### Affected Helm charts

rabbitmq

SeWieland:
### Description of the change

Makes the `/path` paremeter for the scraping endpoint of the rabbitmq serviceMonitor configurable via `.Values.metrics.serviceMonitor.path`

### Benefits

Adds support for configuring more detailed scrape endpoints.

### Possible drawbacks

None, change is non breaking.

### Applicable issues

- fixes #804 

### Additional information

See also <https://www.rabbitmq.com/docs/prometheus#default-endpoint>

### Checklist

- [ ] All commits are signed and verified (required as of January 22, 2026 - see [CONTRIBUTING.md](../CONTRIBUTING.md#setting-up-your-development-environment))

### DIFF
--- a/charts/rabbitmq/Chart.yaml
+++ b/charts/rabbitmq/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rabbitmq
 description: A messaging broker that implements the Advanced Message Queuing Protocol (AMQP)
 type: application
-version: 0.13.2
+version: 0.14.0
 appVersion: "4.2.3"
 keywords:
   - rabbitmq
--- a/charts/rabbitmq/README.md
+++ b/charts/rabbitmq/README.md
@@ -247,17 +247,18 @@ kubectl edit configmap my-rabbitmq-definitions -n <namespace>
 
 ### Metrics configuration
 
-| Parameter                              | Description                                                                                                                 | Default |
-| -------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- | ------- |
-| `metrics.enabled`                      | Enable RabbitMQ metrics (via prometheus plugin)                                                                             | `false` |
-| `metrics.port`                         | RabbitMQ metrics port                                                                                                       | `15692` |
-| `metrics.serviceMonitor.enabled`       | Create ServiceMonitor for Prometheus monitoring                                                                             | `false` |
-| `metrics.serviceMonitor.namespace`     | Namespace for ServiceMonitor                                                                                                | `""`    |
-| `metrics.serviceMonitor.labels`        | Labels for ServiceMonitor                                                                                                   | `{}`    |
-| `metrics.serviceMonitor.annotations`   | Annotations for ServiceMonitor                                                                                              | `{}`    |
-| `metrics.serviceMonitor.interval`      | Scrape interval                                                                                                             | `30s`   |
-| `metrics.serviceMonitor.scrapeTimeout` | Scrape timeout                                                                                                              | `10s`   |
-| `additionalPlugins`                    | Additional RabbitMQ plugins to enable (Prometheus Metrics, PeerDiscoveryK8s and Management plugins are automatically added) | `[]`    |
+| Parameter                              | Description                                                                                                                 | Default    |
+| -------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- | ---------- |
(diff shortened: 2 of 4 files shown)


zOnlyKroks:
One small issue: Your commit is not verified. Whilst this is enforced, I will close both eyes today

SeWieland:
Haha.. I just had to add a new gpg pair to this mail after lunch.. You were too fast ^^
Thanks!

dnijssen:
I'm not sure this will implement my original feature request. This only brings support to change the default path (e.g. `/metrics`) to something else.  

To zoom in a little bit more on the use-case,  I actually still need the (aggregated) metrics from `/metrics` as well, as this tells something about the various nodes in our cluster.  Where as I want to have `/metrics/detailed` to be scraped for queue/exchange specific metrics.  

With the currently implementation I can only change the path from `/metrics` to `/metrics/detailed` , but I still can't have 2 paths in our ServiceMonitor.

Additionally as @erSitzt  stated in https://github.com/CloudPirates-io/helm-charts/issues/804 the `/metrics/detailed` doesn't return anything by default,  you need to specify `vhost` and `family` query parameters .  I'm not sure what will work with the current solution (`metrics.serviceMonitor.path`) as I'm afraid by specifying all in 1 variable will lead to url encoding issues.

SeWieland:
Hmm... sorry, I clearly did not read your issue entirely.
I just migrated over from bitnami and the chart provided the path param (we use /per-object metrics)...
So I saw you issue and referenced it in the PR.

I wasn't aware of your exact use case. I think the additional query params could be added quite easily.
Do you really need multiple scrape endpoints?

dnijssen:
Yes we need 2 endpoints indeed, the `/metrics` returns basic node/cluster level metrics, but not queue specific metrics.  Whereas the `/metrics/detailed` with the appropiate `vhost` and `family` query parameters will return detailed levels of lets say numbe rof messages in a queue.

PR #919, opened 2026-01-28
[common] Fix YAML parse error caused by whitespace control in selectorLabels

spantaleev:
Commit 67b1f4f ("[common] Add support for overriding selector labels (#914)") introduced incorrect Go template whitespace control that causes labels to concatenate on single lines, producing invalid YAML.

Any chart depending on common (etcd, mariadb, valkey, etc.) fails with:

    Error: YAML parse error on etcd/templates/service.yaml: error converting
    YAML to JSON: yaml: line 7: mapping values are not allowed in this context

    helm template my-release oci://registry-1.docker.io/cloudpirates/etcd \
      --version 0.5.0

With --debug, the rendered output shows concatenated labels:

    labels:
      helm.sh/chart: etcd-0.5.0app.kubernetes.io/name: etcdapp.kubernetes.io/instance: my-release

Two uses of `{{-` (leading dash) incorrectly strip newlines:

1. Line 54: `{{- include "cloudpirates.selectorLabels"` strips the newline after `helm.sh/chart`, concatenating it with the first selector label

2. Line 72: `{{- end -}}` strips the newline after `app.kubernetes.io/name`, concatenating it with `app.kubernetes.io/instance`

Remove the leading dash from both locations to preserve the newlines:

- Line 54: `{{- include` -> `{{ include`
- Line 72: `{{- end -}}` -> `{{ end -}}`

### Checklist

### DIFF
--- a/charts/common/Chart.yaml
+++ b/charts/common/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: common
 description: A library chart for common templates and helper functions
 type: library
-version: 2.1.1
-appVersion: "2.1.1"
+version: 2.1.2
+appVersion: "2.1.2"
 home: https://www.cloudpirates.io
 sources:
   - https://github.com/CloudPirates-io/helm-charts/tree/main/charts/common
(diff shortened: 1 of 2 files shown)


zOnlyKroks:
Will upgrade all charts, thank you

PR #923, opened 2026-01-29
[postgres] WAL replication support

mreiche:
### Description of the change

The *Bitnami* chart supported an easy WAL replication setup for Primary -> Standby.
This change brings this feature to this PostgresSQL chart as well.

### Benefits

Makes Primary -> Standby replication setup very easy by enabling:
```yaml
replication:
  enabled: true
```

### Possible drawbacks

- If enabled, it could break custom replication setup done via `customInitScripts` or `initContainers`
- Only works on new databases (without `$PGDATA`) present
- It removes the useless default replication setting in `pg_hba.conf` for localhost restriction only
  - Maybe some users rely on this setting in a specific edge case

### Applicable issues

- None yet

### Additional information

This change also introduces easier `postgresql.conf` configuration like:
```yaml
config:
  postgresql:
    timezone: "UTC"
    locale: "en_US.utf8"
    wal:
      buffers: "16MB"
      level: replica
      maxSenders: 5
      keepSize: 1024
```
which makes these settings deprecated:
```yaml
config:
  postgresqlWalBuffers: ""
  ...
```

**Note**: This *deprecated* setting are used primarly (see `config_test.yml`)

### Checklist

### DIFF
--- a/charts/minio/Chart.yaml
+++ b/charts/minio/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: minio
 description: High Performance Object Storage compatible with Amazon S3 APIs
 type: application
-version: 0.8.1
+version: 0.8.2
 appVersion: "2025.10.15"
 keywords:
   - minio
--- a/charts/minio/templates/bucket-init-configmap.yaml
+++ b/charts/minio/templates/bucket-init-configmap.yaml
@@ -15,55 +15,44 @@ data:
     #!/bin/sh
     set -e
 
-    MINIO_URL="http://{{ include "minio.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
+    MINIO_SERVICE="{{ include "minio.fullname" . }}"
+    MINIO_PORT="{{ .Values.service.port }}"
+    MINIO_URL="http://${MINIO_SERVICE}:${MINIO_PORT}"
     MINIO_STARTUP_TIMEOUT="${MINIO_STARTUP_TIMEOUT:-300}"
 
     echo "=== MinIO Bucket Provisioning ==="
     echo "Target: $MINIO_URL"
+    echo "Service: $MINIO_SERVICE"
+    echo "Port: $MINIO_PORT"
 
     ########################
-    # Check if MinIO is live
-    ########################
-    is_minio_live() {
-        local status_code
-        status_code=$(curl --write-out '%{http_code}' --silent --output /dev/null "${MINIO_URL}/minio/health/live")
-        if [ "$status_code" = "200" ]; then
-            return 0
-        else
-            return 1
-        fi
-    }
-
-    ########################
-    # Wait for MinIO start
+    # Wait for MinIO and configure client
     ########################
     wait_for_minio() {
         local waited_time
         waited_time=0
-        echo "Waiting for MinIO to be ready..."
-        while ! is_minio_live && [ "$waited_time" -lt "$MINIO_STARTUP_TIMEOUT" ]; do
+        echo "Waiting for MinIO to be ready and configuring client..."
+
+        while [ "$waited_time" -lt "$MINIO_STARTUP_TIMEOUT" ]; do
+            # Try to configure alias (might fail if MinIO not ready)
+            if mc alias set local "$MINIO_URL" "${MINIO_ROOT_USER}" "${MINIO_ROOT_PASSWORD}" 2>/dev/null; then
+                echo "Alias configured, testing connection..."
+                # Try a simple operation to verify MinIO is actually ready
+                if mc ls local >/dev/null 2>&1; then
+                    echo "MinIO is ready!"
+                    return 0
+                else
+                    echo "Alias set but MinIO not fully ready yet..."
+                fi
+            fi
+
             echo "MinIO not ready yet, waiting... (${waited_time}s/${MINIO_STARTUP_TIMEOUT}s)"
             sleep 5
             waited_time=$((waited_time + 5))
         done
-        
-        if [ "$waited_time" -ge "$MINIO_STARTUP_TIMEOUT" ]; then
-            echo "Timeout: MinIO not ready after ${MINIO_STARTUP_TIMEOUT}s"
-            return 1
-        fi
-        echo "MinIO is ready!"
-        sleep 2
-    }
 
-    ########################
-    # MinIO client configuration
-    ########################
-    configure_minio_client() {
-        echo "Configuring MinIO client..."
-        mc alias set local "$MINIO_URL" "${MINIO_ROOT_USER}" "${MINIO_ROOT_PASSWORD}"
-        
-        echo "Testing MinIO client connection..."
-        mc ready local
+        echo "Timeout: MinIO not ready after ${MINIO_STARTUP_TIMEOUT}s"
+        return 1
     }
 
     ########################
@@ -106,12 +95,10 @@ data:
     ########################
     # Main execution
     ########################
-    
+
     wait_for_minio
-    
-    configure_minio_client
-    
+
     minio_create_default_buckets
-    
+
     echo "Bucket provisioning completed successfully."
 {{- end }}


mreiche:
Please see that temporary issue.

zOnlyKroks:
> Please see that temporary issue.

Am I too fast or what temporary issue do you reference?

mreiche:
> > Please see that temporary issue.
> 
> Am I too fast or what temporary issue do you reference?

https://github.com/CloudPirates-io/helm-charts/actions/runs/21475219248/job/61857436455?pr=923

zOnlyKroks:
> > > Please see that temporary issue.
> > 
> > 
> > Am I too fast or what temporary issue do you reference?
> 
> https://github.com/CloudPirates-io/helm-charts/actions/runs/21475219248/job/61857436455?pr=923

Fixed ;)

charts/postgres/values.yaml
```diff
@@ -369,3 +386,31 @@ metrics:
...
+    # -- Hostname of the primary server
+    host: ""
+    # -- Port of the primary server
+    port: 5432
+  # -- Replication authentication configuration
+  auth:
+    # -- Username for replication user
+    username: "replication"
+    # -- Password for replication user
+    password: "secret"
```
zOnlyKroks: Could you consider auto generating a empty password, if this field is not set / an empty string?

charts/postgres/values.yaml
```diff
@@ -369,3 +386,31 @@ metrics:
...
+    # -- Use existing secret reference instead of password
+    existingSecret: ""
+    # -- Keys configuration for .existingSecret
+    secretKeys:
+      password: "replication-password"
+  # -- Allow from IP ranges configuration for the primary server
+  allowFrom:
+    # -- Allowed IPv4 network (set empty to disable that feature)
+    ipv4: "0.0.0.0/0"
```
zOnlyKroks, line 414: This is extremely permissive. Could you consider more restrictive defaults like same subnet or cluster CIDR?

charts/postgres/templates/replication-standby-init-configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+    chmod 600 "${pgpass_file}"
+    echo "✅ ${pgpass_file} created"
+
+    if [ -d "${PGDATA}" ]; then
+      echo "${PGDATA} already exists, skipping standby initialization..."
+      exit 0
+    fi
+
+    echo "Initialize database from primary..."
+    PGUSER="${REPLICATION_USER}" PGPASSWORD="${REPLICATION_PASSWORD}" pg_basebackup -h "${primary_host}" -p "${primary_port}" -D "${PGDATA}" --wal-method=stream
```
zOnlyKroks, line 38: Would checking if primary host is reachable first make sense?

charts/postgres/values.yaml
```diff
@@ -369,3 +386,31 @@ metrics:
...
+  # -- Standby server configuration
+  primary:
+    # -- Hostname of the primary server
+    host: ""
+    # -- Port of the primary server
+    port: 5432
+  # -- Replication authentication configuration
+  auth:
+    # -- Username for replication user
+    username: "replication"
```
zOnlyKroks, line 403: Is it ever even checked, that both of these are set and not empty? If so, please do.

zOnlyKroks (changes requested):

charts/postgres/templates/replication-standby-init-configmap.yaml
mreiche: This is something `pg_basebackup` is doing, crashing and restarting the Pod. So it's self healing.

charts/postgres/values.yaml
mreiche: Is there a common Kubernetes CIDR? I could restrict it to `127.0.0.1` instead.

charts/postgres/values.yaml
mreiche: I will take a look onto it.

charts/postgres/values.yaml
mreiche: Ok

charts/postgres/values.yaml
zOnlyKroks: I do not think 127.0.0.1 would work, I will think about if there is a better way or just to leave it as is

mreiche: Usually, you need to configure it anyway and access is usually restricted by additional *NetworkPolicies*

zOnlyKroks: I mean it usually is, good point, but I dont know if this is a best practice we should have in there

charts/postgres/values.yaml
mreiche: It's not useful to generate a password here, because it needs to be the same for both instances. If you have different namespaces or clusters, it will generate two different secrets. I will keep it empty.

zOnlyKroks: > It's not useful to generate a password here, because it needs to be the same for both instances. If you have different namespaces or clusters, it will generate two different secrets. I will keep it empty.

okay, please add a disclaimer / change me note, then we are golden.

charts/postgres/values.yaml
zOnlyKroks: Until this bothers someone, we can leave as is.

zOnlyKroks:
Location: charts/postgres/templates/statefulset.yaml:238-243

`{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) }}` 

to

`{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) .Values.replication.enabled }}`

The condition needs to also include .Values.replication.enabled so the init scripts are mounted when replication is enabled. Currently, the init script seems to never be mounted, thus doesnt run for me

zOnlyKroks:
> Location: charts/postgres/templates/statefulset.yaml:238-243
> 
> `{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) }}`
> 
> to
> 
> `{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) .Values.replication.enabled }}`
> 
> The condition needs to also include .Values.replication.enabled so the init scripts are mounted when replication is enabled. Currently, the init script seems to never be mounted, thus doesnt run for me

Same issue could be for wrapper-script volume mount

mreiche:
> Location: charts/postgres/templates/statefulset.yaml:238-243
> 
> `{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) }}`
> 
> to
> 
> `{{- if or .Values.initdb.scriptsConfigMap .Values.initdb.scripts (and .Values.customUser (or .Values.customUser.name .Values.customUser.existingSecret)) .Values.replication.enabled }}`
> 
> The condition needs to also include .Values.replication.enabled so the init scripts are mounted when replication is enabled. Currently, the init script seems to never be mounted, thus doesnt run for me

Covered by test case

mreiche:
> Same issue could be for wrapper-script volume mount

The wrapper script should not be affected. I dont't get the point of this script anyway.

zOnlyKroks:
> > Same issue could be for wrapper-script volume mount
> 
> The wrapper script should not be affected. I dont't get the point of this script anyway.

`Error: INSTALLATION FAILED: StatefulSet.apps "postgres-test" is invalid: [spec.template.spec.containers[0].volumeMounts[4].name: Not found: "custom-init-scripts", spec.template.spec.containers[0].volumeMounts[5].name: Not found: "wrapper-script"]`

for this test-values.yaml:

replication:
  enabled: true
  auth:
    username: replication
    password: secret123

auth:
  password: postgres123

zOnlyKroks:
> > Same issue could be for wrapper-script volume mount
> 
> The wrapper script should not be affected. I dont't get the point of this script anyway.

irrc it checks for a semaphore file, to see if the init script has run. I do however think that its kind of pointless too

mreiche:
> > > Same issue could be for wrapper-script volume mount
> > 
> > 
> > The wrapper script should not be affected. I dont't get the point of this script anyway.
> 
> `Error: INSTALLATION FAILED: StatefulSet.apps "postgres-test" is invalid: [spec.template.spec.containers[0].volumeMounts[4].name: Not found: "custom-init-scripts", spec.template.spec.containers[0].volumeMounts[5].name: Not found: "wrapper-script"]`
> 
> for this test-values.yaml:
> 
> replication: enabled: true auth: username: replication password: secret123
> 
> auth: password: postgres123

Should be covered now.

zOnlyKroks:
The old wrapper-configmap.yaml doesnt net any value, it would be great if you could remove it.

E.g. delete yaml and from the statefulset:

Lines 77-80 
  {{- else if $shouldMountInitScripts }}
  command:
    - /bin/bash
    - /scripts/postgres-wrapper.sh

Lines 241-242
- name: wrapper-script
  mountPath: /scripts
    
Lines 316-319
name: wrapper-script
    configMap:
      name: {{ printf "%s-wrapper" (include "postgres.fullname" .) }}
      defaultMode: 0755

mreiche:
Could you please merge as squash later.

zOnlyKroks:
> Could you please merge as squash later.

is this urgent? I´d have probably done this tomorrow if it's not urgent.
But yes, we always squash merge

zOnlyKroks:
From a rough look this looks good, give me an half hour or so to take a more in depth look and I will release this today

mreiche:
btw this `zz-mark-init-complete.sh` could also be removed now.

PR #925, opened 2026-01-30
[keycloak]: Allow prepending arguments to keycloak

### LINKED ISSUE #906: [Keycloak] Support for custom keycloak.conf via extraArgsPrefix
### Name and Version

keycloak 0.13.2

I want to start keycloak with a custom `keycloak.conf` which is mounted via secret injection (e.g. Vault Agent or CSI Driver). This allows to securely store the database credentials and other settings in a Vault without creating Kubernetes Secrets.

Right now, it is not possible to configure a path to a custom `keycloak.conf`, because there is only `keycloak.extraArgs` which is inserted at the end of the arguments list.

However, to specify a custom configuration file, it is necessary to add a parameter before the `start` flag:

https://www.keycloak.org/server/configuration#_format_to_include_a_specific_configuration_file

Please add an additional variable `keycloak.extraArgsPrefix` (or similar name) which will be inserted before the `start` flag if present.

Try to overwrite `/opt/keycloak/conf/keycloak.conf` but this is not possible because because of missing write permissions.

Try to append `--config-file` via `extraArgs` but keycloak doesn't start then.

### Additional informations

_No response_

### Affected Helm charts

keycloak

zOnlyKroks:
### Description of the change

Allows prepending arguments to the keycloak startup command

### Benefits

Stated above

### Possible drawbacks

None, by default, its empty

### Applicable issues

- fixes #906

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.13.5
+version: 0.13.6
 appVersion: "26.5.2"
 keywords:
   - keycloak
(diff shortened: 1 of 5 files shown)


FlorianVielhauer:
Looks good. Thank you very much!

PR #954, opened 2026-02-03
[rabbitmq] The management.load_definitions is deprecated and we should use definitions.local.path now

snelis:
### Description of the change

  This PR updates the RabbitMQ configuration to use definitions.local.path instead of the deprecated management.load_definitions directive. The change also adds the required definitions.import_backend = local_filesystem setting.

### Benefits

  - Uses the current RabbitMQ configuration method instead of deprecated approach
  - management.load_definitions has been deprecated by RabbitMQ in favor of definitions.local.path with definitions.import_backend
  - Future-proofs the chart for compatibility with newer RabbitMQ versions

### Possible drawbacks

  This is a drop-in replacement that maintains the same behavior. The configuration change is transparent to users as the Helm values structure remains unchanged.

### Additional information

  The change only affects deployments with definitions.enabled: true. All file paths and volume mounts remain the same - only the RabbitMQ configuration directive names have changed to align with current RabbitMQ best practices.

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.13.6
+version: 0.14.0
 appVersion: "26.5.2"
 keywords:
   - keycloak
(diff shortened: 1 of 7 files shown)


zOnlyKroks:
thank you :)

PR #969, opened 2026-02-05
[ghost]: Allow seperate admin url without second ingress

### LINKED ISSUE #962: Leftover reference to valkey in the ghost chart
Dear maintainers,

https://github.com/CloudPirates-io/helm-charts/blob/52146061d3e2d4620976d52c4b479137eb45adaa/charts/ghost/README.md?plain=1#L28

I think this is a leftover copy&paste error and should reference ghost, not valkey.

### LINKED ISSUE #968: [ghost] Allow setting the admin URL without needing a second ingress
Dear maintainers,

would it be possible to allow setting the admin URL and only fall back to using the URL from the second ingress if the admin URL was not set explicitly?

https://github.com/CloudPirates-io/helm-charts/blob/main/charts/ghost/templates/configmap.yaml#L56

Using helm's default function should allow this:
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

Something like this:
```
"url": {{ .Values.admin_url | default(printf "https://%s" (index .Values.ingress.hosts 1).host) | quote }}
```

This would allow more flexibility, for instance if a user does not want to use the ingress or if the admin url should not be a separate URL.

Thanks in advance!

Kind Regards,
Johannes

zOnlyKroks:
### Description of the change

Adds the ability to specify a separate admin url, that defaults back to the second ingress url

### Benefits

Ability to easily specify a different admin url

### Possible drawbacks

None, change is backwards compatible

### Applicable issues

- fixes #968
- fixes #962

### Checklist

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.14.2
+version: 0.14.3
 appVersion: "26.5.2"
 keywords:
   - keycloak
--- a/charts/keycloak/values.yaml
+++ b/charts/keycloak/values.yaml
@@ -437,7 +437,7 @@ initContainers:
     ## @param initContainers.waitForPostgres.repository PostgreSQL image repository
     repository: postgres
     ## @param initContainers.waitForPostgres.tag PostgreSQL image tag
-    tag: "18.1@sha256:3bb77a07b9ce8b8de0fe6d9b063adf20b9cca1068a1bcdc054cac71a69ce0ca6"
+    tag: "18.1@sha256:1090bc3a8ccfb0b55f78a494d76f8d603434f7e4553543d6e807bc7bd6bbd17f"
     ## @param initContainers.waitForPostgres.pullPolicy PostgreSQL image pull policy
     pullPolicy: IfNotPresent
     ## @param initContainers.waitForPostgres.resources Resource requests and limits for PostgreSQL init container
--- a/charts/nginx/Chart.yaml
+++ b/charts/nginx/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: nginx
 description: Nginx is a high-performance HTTP server and reverse proxy.
 type: application
-version: 0.5.9
-appVersion: "1.29.4"
+version: 0.5.10
+appVersion: "1.29.5"
 keywords:
   - nginx
   - web server
--- a/charts/nginx/values.yaml
+++ b/charts/nginx/values.yaml
@@ -24,7 +24,7 @@ image:
   ## @param image.repository Nginx image repository
   repository: nginx
   ## @param image.tag Nginx image tag
-  tag: "1.29.4-alpine@sha256:4870c12cd2ca986de501a804b4f506ad3875a0b1874940ba0a2c7f763f1855b2"
+  tag: "1.29.5-alpine@sha256:05fcc6f6c80dc65c0f241f250677aad26e0962d118abaa65a513796862ff915d"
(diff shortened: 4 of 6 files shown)


charts/ghost/README.md
```diff
@@ -64,9 +61,20 @@ To verify the helm chart before installation, copy the public key to the file `c
 cosign verify --key cosign.pub registry-1.docker.io/cloudpirates/ghost:<version>
 ```
 
-
 ## Configuration
 
+### Custom Admin URL
+
+By default, Ghost expects a second ingress host for the admin interface. If you want to customize the admin URL or use a different configuration, you can set it explicitly:
```
kastl-ars: `By default, this chart expects a second ingress`

kastl-ars:
Looks good to me, I'll try to give it a spin later.

kastl-ars:
Unfortunately, this does not work.

1. The ingress host is also referenced in the `templates/NOTES.txt` file of the ghost chart.
2. Helm executes the `index` function even if `config.admin.url` is set.

The best solution that comes to my mind now is a helper function. Something like this worked locally

```
{{/*
Admin URL
*/}}
{{- define "ghost.admin_url" -}}
{{- if .Values.config.admin.url }}
{{- .Values.config.admin.url }}
{{- else }}
{{- printf "https://%s" (index .Values.ingress.hosts 1) }}
{{- end }}
{{- end }}
```

configmap:
```
      "admin": {
        "url": {{ include "ghost.admin_url" . | quote }}
      },
```

```
helm template my-ghost -f my-values.yaml .
```

Using these values in `my-values.yaml`:

```
config:
  admin:
    url: "https://ghost.example.org"
ingress:
  enabled: false
```

zOnlyKroks:
> Unfortunately, this does not work.
> 
> 1. The ingress host is also referenced in the `templates/NOTES.txt` file of the ghost chart.
> 2. Helm executes the `index` function even if `config.admin.url` is set.
> 
> The best solution that comes to my mind now is a helper function. Something like this worked locally
> 
> ```
> {{/*
> Admin URL
> */}}
> {{- define "ghost.admin_url" -}}
> {{- if .Values.config.admin.url }}
> {{- .Values.config.admin.url }}
> {{- else }}
> {{- printf "https://%s" (index .Values.ingress.hosts 1) }}
> {{- end }}
> {{- end }}
> ```
> 
> configmap:
> 
> ```
>       "admin": {
>         "url": {{ include "ghost.admin_url" . | quote }}
>       },
> ```
> 
>  ```
> helm template my-ghost -f my-values.yaml .
> ```
> 
> Using these values in `my-values.yaml`:
> 
> ```
> config:
>   admin:
>     url: "https://ghost.example.org"
> ingress:
>   enabled: false
> ```

you may look again

charts/ghost/templates/_helpers.tpl
```diff
@@ -63,4 +63,15 @@ Create the name of the service account to use
...
+{{- end }}
+
+{{/*
+Return the admin URL for Ghost
+*/}}
+{{- define "ghost.admin_url" -}}
+{{- if .Values.config.admin.url }}
+{{- .Values.config.admin.url }}
+{{- else }}
+{{- printf "https://%s" (index .Values.ingress.hosts 1).host }}
```
kastl-ars, line 75: Maybe it would be a good idea to catch a missing second ingress by adding another if-condition and printing a message for the enduser, if there is no second ingress?

kastl-ars:
Looks good to me. Thanks for taking care so fast... :-)

charts/ghost/templates/_helpers.tpl
kastl-ars: Thanks for implementing this, @zOnlyKroks

kastl-ars (approved):
Looks good to me, works in my local tests (with `helm template`).

PR #972, opened 2026-02-05
[ghost]: Move secrets from configmap to secret

### LINKED ISSUE #970: [ghost] Allow setting database and STMP credentials from Kubernetes secrets
### Name and Version

ghost

Dear maintainers,

I would really like to avoid having credentials in my helm values. Current best practice is to reference a Kubernetes secret that can be safely created, while the helm values might be stored in Git (think GitOps).

Unfortunately the ghost chart sets the complete ghost configuration in the values directly.

Would it be possible to allow using an external secret for the credentials?

```
config:
  database:
    client: "mysql"
    externalConnection:
      host: "ghost-mariadb"
      port: 3306
      # do NOT set the credentials directly
      # user: "ghost"
      # password: "changeme"
      externalSecret:
        name: my-ghost-credentials
        usernameKey: username
        passwordKey: password
      database: "ghost"
```

This would reference a secret called `my-ghost-credentials` that contains both a `username` and a `password` key.

Of course, the real magic would be in the background, to reference the external secret and make its contents available in the ghost pod. Currently the config is stored in a configMap, which prevents this whole idea.

What do you think?

Kind Regards,
Johannes

_No response_

### Additional informations

_No response_

### Affected Helm charts

_No response_

### LINKED ISSUE #971: [ghost] Use a secret for the configuration, not a configMap (as it contains credentials)
### Name and Version

ghost

The current state of the chart creates a configMap containing the complete ghost configuration. This includes credentials like the database or SMTP passwords, e.g.

https://github.com/CloudPirates-io/helm-charts/blob/main/charts/ghost/templates/configmap.yaml#L23

#970 proposes referencing secrets for those credentials, but needs a lot of work.

In the meantime it would be best to convert the configMap into a Kubernetes secret, to at least have some of the security-relevant advantages of Kubernetes secrets (not being stored on disk, etc. pp.). Of course, the source values will still be in plaintext in the helm values, but at least inside Kubernetes we get a little more security.

Kind Regards,
Johannes

_No response_

The ghost configuration is in a secret.

The ghost configuration is in a configMap.

### Additional informations

_No response_

### Affected Helm charts

ghost

zOnlyKroks:
### Description of the change

Gives the user ability to move secrets out of the configmap, into a dedicated secret.

### Benefits

Better security / best practices

### Possible drawbacks

None, its backwards compatible and optional

### Applicable issues

- fixes #970 
- fixes #971

### Checklist

### DIFF
--- a/charts/ghost/README.md
+++ b/charts/ghost/README.md
@@ -132,7 +132,7 @@ config:
 
 **For bundled MariaDB:**
 
-You can also use an existing secret with the bundled MariaDB deployment. Create a secret:
+You can also use an existing secret with the bundled MariaDB deployment. Create a secret with the required keys:
 
 ```bash
 kubectl create secret generic ghost-mariadb-credentials \
@@ -147,11 +147,12 @@ mariadb:
   enabled: true
   auth:
     database: ghost
-    existingSecret:
-      name: "ghost-mariadb-credentials"
-      usernameKey: "username"
-      passwordKey: "mariadb-password"
+    existingSecret: "ghost-mariadb-credentials"
 ```
+
+**Note:** The secret must contain the following keys:
+- `username`: database username
+- `mariadb-password`: database password for the user
 
 #### SMTP Credentials from Secret
 
@@ -237,16 +238,14 @@ The following tables list the configurable parameters of the Ghost chart organiz
 
 ### Database Parameters
 
-| Parameter                                                 | Description                                        | Default              |
-| --------------------------------------------------------- | -------------------------------------------------- | -------------------- |
-| `mariadb.enabled`                                         | Deploy MariaDB as dependency                       | `true`               |
-| `mariadb.auth.database`                                   | MariaDB database name                              | `ghost`              |
-| `mariadb.auth.username`                                   | MariaDB username                                   | `ghost`              |
-| `mariadb.auth.password`                                   | MariaDB password                                   | `changeme`           |
-| `mariadb.auth.existingSecret.name`                        | Name of existing secret with MariaDB credentials   | `""`                 |
-| `mariadb.auth.existingSecret.usernameKey`                 | Key in secret containing database username         | `"username"`         |
-| `mariadb.auth.existingSecret.passwordKey`                 | Key in secret containing database password         | `"mariadb-password"` |
-| `mariadb.auth.allowEmptyRootPassword`                     | Allow empty root password                          | `false`              |
+| Parameter                                                 | Description                                        | Default    |
+| --------------------------------------------------------- | -------------------------------------------------- | ---------- |
+| `mariadb.enabled`                                         | Deploy MariaDB as dependency                       | `true`     |
+| `mariadb.auth.database`                                   | MariaDB database name                              | `ghost`    |
+| `mariadb.auth.username`                                   | MariaDB username                                   | `ghost`    |
+| `mariadb.auth.password`                                   | MariaDB password                                   | `changeme` |
+| `mariadb.auth.existingSecret`                             | Name of existing secret with MariaDB credentials (must contain keys: username, mariadb-password) | `""` |
+| `mariadb.auth.allowEmptyRootPassword`                     | Allow empty root password                          | `false`    |
 | `config.database.externalConnection.existingSecret.name`  | Name of existing secret for external DB credentials| `""`       |
 | `config.database.externalConnection.existingSecret.usernameKey` | Key in secret containing database username   | `username` |
 | `config.database.externalConnection.existingSecret.passwordKey` | Key in secret containing database password   | `password` |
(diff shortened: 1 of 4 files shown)


charts/ghost/templates/configmap.yaml
```diff
@@ -28,8 +28,13 @@ data:
         "connection": {
           "host": {{ .Values.config.database.externalConnection.host | quote }},
           "port": {{ .Values.config.database.externalConnection.port }},
+          {{- if .Values.config.database.externalConnection.existingSecret.name }}
+          "user": "",
+          "password": "",
+          {{- else }}
           "user": {{ .Values.config.database.externalConnection.user | quote }},
```
kastl-ars: Instead of keeping the credentials in the config file, I would make them environment variables in both cases, this would solve #971 as the configMap would no longer contain credentials.
What do you think?

charts/ghost/templates/configmap.yaml
```diff
@@ -28,8 +28,13 @@ data:
         "connection": {
           "host": {{ .Values.config.database.externalConnection.host | quote }},
           "port": {{ .Values.config.database.externalConnection.port }},
+          {{- if .Values.config.database.externalConnection.existingSecret.name }}
+          "user": "",
```
kastl-ars, line 31: I am not familiar with the Ghost internals, does it need an empty value to read this from environment variables?

zOnlyKroks:
@kastl-ars you may now look again

charts/ghost/templates/deployment.yaml
```diff
@@ -73,12 +73,64 @@ spec:
...
+                secretKeyRef:
+                  name: {{ .Values.config.mail.options.auth.existingSecret.name }}
+                  key: {{ .Values.config.mail.options.auth.existingSecret.passKey }}
+            {{- else }}
+            - name: mail__options__auth__user
+              value: {{ .Values.config.mail.options.auth.user | quote }}
+            - name: mail__options__auth__pass
+              value: {{ .Values.config.mail.options.auth.pass | quote }}
+            {{- end }}
             {{- with .Values.extraEnvVars }}
```
kastl-ars: I think the following line got lost but is still needed (before this one, indented two spaces more):

```
{{- if .Values.extraEnvVars }}
```

kastl-ars: This also means the delete end line needs to be added back, with two more spaces

charts/ghost/values.yaml
```diff
@@ -213,6 +213,15 @@ config:
       user: "ghost"
       password: "changeme"
       database: "ghost"
+      ## @param config.database.externalConnection.existingSecret Use existing secret for database credentials
```
kastl-ars, line 216: I think you need to add the `existingSecret` part for mariadb, too?

You have referenced `.Values.mariadb.auth.existingSecret` in the templates.

charts/ghost/README.md
```diff
@@ -89,6 +89,66 @@ config:
...
+Create a Kubernetes secret with your database credentials:
+
+```bash
+kubectl create secret generic ghost-db-credentials \
+  --from-literal=username=ghost \
+  --from-literal=password=your-secure-password
+```
+
+Reference this secret in your `my-values.yaml`:
+
```
kastl-ars, line 115: As this is also possible with the mariadb, can you please add another example or at least a statement that this also works with `mariadb.enabled`?

charts/ghost/templates/deployment.yaml
```diff
@@ -73,12 +73,61 @@ spec:
               containerPort: {{ .containerPort }}
               protocol: {{ .protocol | default "TCP" }}
             {{- end }}
-          {{- if .Values.extraEnvVars }}
           env:
+            {{- if .Values.mariadb.enabled }}
```
kastl-ars, line 95: Is it intentional that the database user cannot be set from a secret when using the mariadb subchart?

charts/ghost/values.yaml
kastl-ars: This is still missing, isn't it? Resolved by accident? Forgotten to push?

charts/ghost/README.md
```diff
@@ -97,6 +97,89 @@ config:
...
+```
+
+Reference it in your `my-values.yaml`:
+
+```yaml
+mariadb:
+  enabled: true
+  auth:
+    database: ghost
+    existingSecret: "ghost-mariadb-credentials"
```
kastl-ars: Did you leave out the username and password keys for brevity?

zOnlyKroks: yes

kastl-ars: Then I would at least add `...` in the next line, to show that this is not a full example. What do you think?

zOnlyKroks: > Then I would at least add `...` in the next line, to show that this is not a full example. What do you think?

i will probably extend it later today

zOnlyKroks:
@kastl-ars I presume that fits your needs?

charts/ghost/values.yaml
kastl-ars: This is still missing

kastl-ars:
> @kastl-ars I presume that fits your needs?

> .Values.mariadb.auth.existingSecret

The `existingSecret` part in the internal mariadb section of the `values.yaml` is still missing, isn't it?

PR #1044, opened 2026-02-23
[rabbitmq]: Auto-inject missing vhosts and auth user into definitions to prevent boot failure

### LINKED ISSUE #1024: RabbitMQ Crash when using auth and definitions.permissions together
### Name and Version

rabbitmq

**Deploy RabbitMQ with:**
```yaml
auth:
  enabled: true
  username: myuser
  existingSecret: my-secret
definitions:
  enabled: true
  permissions:
    - user: myuser
      vhost: /myuser
      configure: ".*"
      write: ".*"
      read: ".*"
```

**Expected Behavior**
- The user myuser should be created by auth (via RABBITMQ_DEFAULT_USER/RABBITMQ_DEFAULT_PASS environment variables)
- Then definitions.permissions should grant the user permissions on the vhost

_No response_

_No response_

When both auth and definitions with permissions are enabled, RabbitMQ crashes on startup with the error:
`{exit,{18431,{error,<<"{no_such_user,<<\"myuser\">>}">>}},...}
Runtime terminating during boot`

### Additional informations

_No response_

### Affected Helm charts

rabbitmq

zOnlyKroks:
### Description of the change

When auth.enabled is true and definitions.permissions references the default user or a vhost not explicitly defined in definitions.vhosts, RabbitMQ crashes during boot. This change updates the definitions ConfigMap template to:

Auto-inject the auth user into users if not already explicitly defined
Auto-inject any vhosts referenced in permissions that are not already explicitly defined in vhosts

### Benefits

RabbitMQ no longer crashes on boot when using auth.enabled together with definitions.permissions
Users no longer need to redundantly declare the default auth user or referenced vhosts in the definitions block — the chart handles it automatically
Reduces misconfiguration surface area with no breaking changes to existing values

### Possible drawbacks

The auto-injected auth user entry has an empty password field and administrator tag hardcoded. If a user wants to define the default user with different tags via definitions.users, they must do so explicitly — the injection is skipped if the username is already present
Does not cover existingConfigMap or existingSecret definitions sources — users managing their own definitions payload are still responsible for including the correct users and vhosts

### Applicable issues

- fixes #1024

### Additional information

RabbitMQ loads definitions atomically at boot before RABBITMQ_DEFAULT_USER/RABBITMQ_DEFAULT_PASS env vars are applied, so the user referenced in permissions must exist within the definitions payload itself. Similarly, vhosts must be defined before permissions referencing them are applied, as RabbitMQ processes definitions in a fixed order: users → vhosts → permissions.

### Checklist

### DIFF
--- a/.github/workflows/changelog-and-release.yaml
+++ b/.github/workflows/changelog-and-release.yaml
@@ -58,7 +58,7 @@ jobs:
       has_changes: ${{ steps.generate-changelog.outputs.has_changes }}
     steps:
       - name: Checkout main branch
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
           token: ${{ secrets.CHANGELOG_PAT || secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


timhovius:
I am experiencing the same issue. When custom definitions are loaded, I am unable to log in to the management interface.

zOnlyKroks:
> I am experiencing the same issue. When custom definitions are loaded, I am unable to log in to the management interface.

does this pr solve your issue or does it remain?

PR #1091, opened 2026-03-04
[redis] Added Option to run redis as Deployment

### LINKED ISSUE #1019: Allow using Deployment instead of StatefulSet when persistence.enabled=false
### Name and Version

redis

We experienced a Redis master pod being stuck in Terminating for over 26 hours after a node failure, causing a complete Redis outage for the affected application. The only way to recover was to manually force-delete the pod via ArgoCD. Every other workload was automatically rescheduled to healthy nodes within seconds — Redis was the only service that remained down.

We expect this to be due to the Statefulset deployment. When a Kubernetes node fails unexpectedly, pods managed by a StatefulSet get stuck in Terminating state indefinitely. This happens because the StatefulSet controller enforces a strict identity guarantee — it will never create a replacement pod until the old one is confirmed deleted by the kubelet. If the node is dead, the kubelet never responds, resulting in a deadlock.

This does exist for good reason: preventing two pods from writing to the same PersistentVolume simultaneously, which could cause data corruption.

However, when persistence.enabled=false (as in our case), Redis runs entirely on emptyDir volumes. There is no persistent state to protect. In this case, the StatefulSet's strict identity guarantee provides zero benefit while actively harming availability - the Redis pod cannot be rescheduled until the dead node is manually cleaned up (e.g. via kubectl delete node, force-deleting the pod, or applying an out-of-service taint).

When persistence.enabled is set to false, the chart should deploy Redis as a Deployment (with strategy.type: Recreate) instead of a StatefulSet. Since there are no PersistentVolumeClaims and no stable storage to protect, there is no reason to use a StatefulSet.

Alternatively, a separate value like useStatefulSet: true/false could be introduced to give users explicit control, defaulting to true when persistence.enabled=true and false when persistence.enabled=false.

We're force-deleted Pods after a Node failure to have them rescheduled

### Additional informations

Environment: k3s, Helm chart redis-20.2.2, Redis 7.4.1, no persistence, no replicas.

### Affected Helm charts

redis

sschmidt:
### Description of the change

- solves https://github.com/CloudPirates-io/helm-charts/issues/1019 

### Benefits

Allows to use deployments instead of statefulsets when running redis without persistence.

### Possible drawbacks

If `useDeployment` is enabled, we don't have static dns for redis nodes (e.g. redis-node-0-headless.svc.local ...). This shouldn't be a problem as most people will use the provided services (master/headless/sentinel) to connect. Anyway, as this wouldn't be backwards compatible we decided to only enable deployment's when this new values is set instead of doing it by default wenn `persistence.enabled` is false.

### Applicable issues

- fixes #1019 

### Checklist

### DIFF
--- a/charts/redis/Chart.yaml
+++ b/charts/redis/Chart.yaml
@@ -42,9 +42,7 @@ annotations:
       url: https://www.cloudpirates.io
   artifacthub.io/changes: |-
     - kind: changed
-      description: "[oliver006/redis_exporter] Update image to v1.82.0"
+      description: "Properly set the default for existingSecretPasswordKey inside the values.yaml"
       links:
-        - name: "Pull Request"
-          url: "https://github.com/CloudPirates-io/helm-charts/pull/1100"
         - name: "Changelog"
           url: "https://github.com/CloudPirates-io/helm-charts/blob/main/charts/redis/CHANGELOG.md"
--- a/.github/workflows/changelog-and-release.yaml
+++ b/.github/workflows/changelog-and-release.yaml
@@ -548,7 +548,14 @@ jobs:
           fi
 
           # Normal detection logic
-          for chart_dir in charts/*; do
+          # If a specific chart was requested via workflow_dispatch, only process that chart
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ] && [ -n "${{ github.event.inputs.chart_name }}" ]; then
+            CHART_DIRS="charts/${{ github.event.inputs.chart_name }}"
+          else
+            CHART_DIRS="charts/*"
+          fi
+
+          for chart_dir in $CHART_DIRS; do
             [ ! -f "$chart_dir/Chart.yaml" ] && continue
 
             CHART_NAME=$(basename "$chart_dir")
--- a/charts/nginx/CHANGELOG.md
+++ b/charts/nginx/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this chart will be documented in this file.
 
+## [0.8.0] - 2026-03-05
+
+- Add daemonset mode (#1092) ([3cea25af](https://github.com/CloudPirates-io/helm-charts/commit/3cea25af))
+
 ## [0.7.0] - 2026-02-25
 
 - [nginx]: existingServerConfigConfigmap volume mismatch fix (#1059) ([5dd64454](https://github.com/CloudPirates-io/helm-charts/commit/5dd64454))
--- a/charts/nginx/Chart.yaml
+++ b/charts/nginx/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: nginx
 description: Nginx is a high-performance HTTP server and reverse proxy.
 type: application
-version: 0.8.0
+version: 0.8.1
 appVersion: "1.29.5"
 keywords:
   - nginx
@@ -42,9 +42,9 @@ annotations:
       url: https://www.cloudpirates.io
   artifacthub.io/changes: |-
     - kind: added
-      description: "Add daemonset mode"
+      description: "Add internalTrafficPolicy to service"
       links:
         - name: "Pull Request"
-          url: "https://github.com/CloudPirates-io/helm-charts/pull/1092"
+          url: "https://github.com/CloudPirates-io/helm-charts/pull/1103"
         - name: "Changelog"
           url: "https://github.com/CloudPirates-io/helm-charts/blob/main/charts/nginx/CHANGELOG.md"
--- a/charts/nginx/README.md
+++ b/charts/nginx/README.md
@@ -4,7 +4,7 @@
 
 # Nginx
 
-nginx ("engine x") is an HTTP web server, reverse proxy, content cache, load balancer, TCP/UDP proxy server, and mail proxy server. 
+nginx ("engine x") is an HTTP web server, reverse proxy, content cache, load balancer, TCP/UDP proxy server, and mail proxy server.
 
 ## Installing the Chart
 
@@ -62,7 +62,7 @@ serverConfig: |
     listen 0.0.0.0:8080;
     root /usr/share/nginx/html;
     index index.html index.htm;
-    
+
     location / {
       try_files $uri $uri/ /index.html;
     }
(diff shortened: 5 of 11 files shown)


charts/redis/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: redis
 description: An open source, in-memory data structure store used as a database, cache, and message broker.
 type: application
-version: 0.25.6
+version: 0.25.7
```
zOnlyKroks: I would propose 0.26.0

charts/redis/templates/statefulset.yaml
```diff
@@ -467,6 +691,223 @@ spec:
...
+                CURRENT_ROLE=$(redis-cli {{- if .Values.auth.enabled }} -a "${REDIS_PASSWORD}"{{- end }} --no-auth-warning -h "${REDIS_HOST}" -p {{ .Values.service.port }} role 2>/dev/null | head -1)
+                if [ "$MASTER_HOST" = "$MY_IP" ] && [ "$CURRENT_ROLE" != "master" ]; then
+                  echo "Promoting local Redis to master (REPLICAOF NO ONE)..."
+                  redis-cli {{- if .Values.auth.enabled }} -a "${REDIS_PASSWORD}"{{- end }} --no-auth-warning -h "${REDIS_HOST}" -p {{ .Values.service.port }} REPLICAOF NO ONE 2>/dev/null && echo "Local Redis promoted to master" || echo "Warning: REPLICAOF NO ONE failed"
+                elif [ "$MASTER_HOST" != "$MY_IP" ] && [ "$CURRENT_MASTER" != "$MASTER_HOST" ]; then
+                  echo "Reconfiguring local Redis to replicate from master ${MASTER_HOST} (was: ${CURRENT_MASTER:-none})..."
+                  redis-cli {{- if .Values.auth.enabled }} -a "${REDIS_PASSWORD}"{{- end }} --no-auth-warning -h "${REDIS_HOST}" -p {{ .Values.service.port }} REPLICAOF "$MASTER_HOST" {{ .Values.service.port }} 2>/dev/null && echo "Local Redis now replicating from $MASTER_HOST" || echo "Warning: REPLICAOF failed"
```
zOnlyKroks: I think the option for TLS is missing here, e.g. 
`{{ if .Values.tls.enabled }}{{ .Values.tls.port }}{{ else }}{{ .Values.service.port }}{{ end }}` and so on

charts/redis/templates/statefulset.yaml
```diff
@@ -11,7 +16,16 @@ metadata:
 {{- include "cloudpirates.tplvalues.render" (dict "value" (include "redis.annotations" .) "context" .) | indent 4 }}
   {{- end }}
 spec:
+  {{- if $useDeployment }}
+  strategy:
+    {{- if or (eq .Values.architecture "standalone") (not .Values.sentinel.enabled) }}
+    type: Recreate
```
zOnlyKroks: Maybe a less destructive RollingUpdate?

charts/redis/templates/statefulset.yaml
```diff
@@ -60,6 +74,215 @@ spec:
...
+              echo "min-slaves-max-lag 0" >> /tmp/redis.conf
+
+              # Sentinel discovery: retry loop across all headless IPs, skipping self
+              CURRENT_MASTER=""
+              SENTINEL_FOUND=false
+              echo "Searching for existing sentinel..."
+              for retry in $(seq 1 5); do
+                for SENTINEL_IP in $(getent hosts "${HEADLESS_SVC}" 2>/dev/null | awk '{print $1}' | sort -u); do
+                  [ "$SENTINEL_IP" = "$MY_IP" ] && continue
+                  MASTER_INFO=$(redis-cli -h "${SENTINEL_IP}" -p {{ .Values.sentinel.port }} {{- if .Values.auth.sentinel }} -a "${REDIS_PASSWORD}"{{- end }} sentinel get-master-addr-by-name {{ .Values.sentinel.masterName }} 2>/dev/null | head -1 || echo "")
```
zOnlyKroks: is the --no-auth-warning needed like the other redis-cli usages?

charts/redis/templates/statefulset.yaml
```diff
@@ -60,6 +74,215 @@ spec:
...
+              MY_IP=$(hostname -i | grep -oE '([0-9a-f:]+:+)+[0-9a-f]+' | head -1)
+              if [ -z "$MY_IP" ]; then
+                echo "Warning: No IPv6 address found, falling back to first available IP"
+                MY_IP=$(hostname -i | awk '{print $1}')
+              fi
+              {{- else }}
+              MY_IP=$(hostname -i | awk '{print $1}')
+              {{- end }}
+              echo "Using IP address: ${MY_IP}"
+              echo "replica-announce-ip ${MY_IP}" >> /tmp/redis.conf
```
zOnlyKroks, line 210: duplicate see line 211

charts/redis/templates/statefulset.yaml
```diff
@@ -1,6 +1,11 @@
 {{- include "redis.auth.acl.validate" . -}}
+{{- $useDeployment := and .Values.useDeployment (not .Values.persistence.enabled) -}}
 apiVersion: apps/v1
+{{- if $useDeployment }}
```
zOnlyKroks, line 5: The file is still named statefulset.yaml but now renders either StatefulSet or Deployment — a comment at the top would help future maintainers.

charts/redis/Chart.yaml
sschmidt: done

charts/redis/templates/statefulset.yaml
sschmidt: done

charts/redis/templates/statefulset.yaml
sschmidt: RollingUpdate isn't a good choice for replication deployment mode: the headless service returns stale IPs of terminated pods during rollout. The init container's lowest-IP master election then picks a dead IP and blocks on a 60s timeout (downtime). This leaves the cluster without a master and causes 2-3 sentinel restarts from s_down liveness failures.

charts/redis/templates/statefulset.yaml
sschmidt: Not needed, but added for consistency and reduced logs.

charts/redis/templates/statefulset.yaml
sschmidt: done

charts/redis/templates/statefulset.yaml
sschmidt: done

PR #1093, opened 2026-03-04
[mongodb] fix a race condition between detect_cluster and Kubernetes Service DNS

lucendio:
This change set addresses three issues:

1. attempting to detect a cluster only after the Service DNS resolves
2. properly detecting whether there is an existing replicaset or not
3. working around the chicken/egg problem of the DNS record of the
   Service not being available before the pod(s) enter ready state

The creation of the Service is an asynchronous operation. As a result,
the subsequent mongoshell command may fail, but for a different reason
than intended:

```
MongoNetworkError: getaddrinfo ENOTFOUND <headless-service-name>.<namespace>.svc.cluster.local
```

Checking the DNS availability upfront, solves this issue. Please note,
that as the Service is always rendered and created, there is no need
to limit the amount of iterations of the `until` loop.

Unfortunately, as the pod(s) are not ready at this point (extra-init.sh
did not finished yet), the DNS record in question is not yet created by
Kubernetes. To work around this default behaviour, the immediate
publication of the record is enforced by adding the respective flag to
the ServiceSpec.

Finally, adding this DNS check, makes the `db.version()` check
useless as it's not an appropriate way of finding out whether a
replicaset already exists or not. Which is way, the mongoshell command
has been changed.

### DIFF
--- a/charts/mongodb/Chart.yaml
+++ b/charts/mongodb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mongodb
 description: MongoDB a flexible NoSQL database for scalable, real-time data management
 type: application
-version: 0.12.4
+version: 0.12.5
 appVersion: "8.2.5"
 keywords:
   - mongodb


lucendio:
Well, this wont quite work. The DNS entry of the Service is only available, after (!) the referred pods become *ready*. And they only become healthy after the `extra-init.sh` ran successfully.

The race condition becomes a chicken/egg situation.

Adding `publishNotReadyAddresses: true` to the Spec of the headless service would resolve the chicken/egg situation. Is this additional change feasible?

zOnlyKroks:
> Well, this wont quite work. The DNS entry of the Service is only available, after (!) the referred pods become _ready_. And they only become healthy after the `extra-init.sh` ran successfully.
> 
> The race condition becomes a chicken/egg situation.
> 
> Adding `publishNotReadyAddresses: true` to the Spec of the headless service would resolve the chicken/egg situation. Is this additional change feasible?

Looking at the sharded cluster headless services, service-configserver.yaml and service-shard.yaml both already have publishNotReadyAddresses true. So it is a viable change

lucendio:
> > Well, this wont quite work. The DNS entry of the Service is only available, after (!) the referred pods become _ready_. And they only become healthy after the `extra-init.sh` ran successfully.
> > The race condition becomes a chicken/egg situation.
> > Adding `publishNotReadyAddresses: true` to the Spec of the headless service would resolve the chicken/egg situation. Is this additional change feasible?
> 
> Looking at the sharded cluster headless services, service-configserver.yaml and service-shard.yaml both already have publishNotReadyAddresses true. So it is a viable change

Great. I just made the additional change and updated the description.

zOnlyKroks:
> > > Well, this wont quite work. The DNS entry of the Service is only available, after (!) the referred pods become _ready_. And they only become healthy after the `extra-init.sh` ran successfully.
> > > The race condition becomes a chicken/egg situation.
> > > Adding `publishNotReadyAddresses: true` to the Spec of the headless service would resolve the chicken/egg situation. Is this additional change feasible?
> > 
> > 
> > Looking at the sharded cluster headless services, service-configserver.yaml and service-shard.yaml both already have publishNotReadyAddresses true. So it is a viable change
> 
> Great. I just made the additional change and updated the description.

Mind version bumping this again?

zOnlyKroks:
@lucendio Mind fixing your commits beeing unverified? Verified commits are required as of late before merging, thank you

lucendio:
> @lucendio Mind fixing your commits beeing unverified? Verified commits are required as of late before merging, thank you

done

zOnlyKroks:
I dont think all of these changes should be here

lucendio:
> I dont think all of these changes should be here

well, according to the email notifications, you were the one merging `main` into `patch-1` multiple times since this PR was created.

zOnlyKroks:
> > I dont think all of these changes should be here
> 
> well, according to the email notifications, you were the one merging `main` into `patch-1` multiple times since this PR was created.

interesting, could be that I miss clicked whilst managing other pr´s, I am sorry for the caused inconvenience

lucendio:
> > > I dont think all of these changes should be here
> > 
> > 
> > well, according to the email notifications, you were the one merging `main` into `patch-1` multiple times since this PR was created.
> 
> interesting, could be that I miss clicked whilst managing other pr´s, I am sorry for the caused inconvenience

no worries. I dont know what happened on GH's side. but, I just learned that GH does not automatically update the main branch of my fork. I had to click the *sync fork* button at the top of my fork. Afterwards the `origin/main` rebase actually did what it was supposed to. I think, we should be good now.

PR #1163, opened 2026-03-18
[postgres]: Allow for configuration of wal settings without replication

### LINKED ISSUE #1160: Postgres: `max_wal_senders` via `config.postgresql` silently dropped if no replication is configured
### Name and Version

postgres 0.17.2

```yaml
# values.yaml excerpt
image:
  repository: postgres
  tag: "17.5"

config:
  postgresql:
    wal_level: minimal
    # Necessary for switching wal_level to 'minimal'
    max_wal_senders: 0
...
```

```bash
helm upgrade postgres oci://registry-1.docker.io/cloudpirates/postgres \
  --version 0.17.2 \
  --install \
  --wait \
  --atomic \
  --namespace databases-qa \
  --values values.yaml \
  --set-string auth.password=${PGPASSWORD}
```

... see the Postgres logs:
```log
│ 2026-03-17 14:14:02.474 CET [1] FATAL:  WAL streaming ("max_wal_senders" > 0) requires "wal_level" to be "replica" or "logical" 
```

_No response_

There are use cases for deploying Postgres without replication but still having `max_wal_senders` set to `0`.

For instance, when using a Postgres instance for testing purposes that doesn't need persistence, replication, WAL, etc. Here setting the `config.postgresql.max_wal_senders` to `0` for performance reasons is a valid case.

The chart offers `config.postgresql.max_wal_senders` but silently drops its value when deployed without replication; instead one has to use `config.extraConfig` to set `max_wal_senders` to `0`. 

### Additional informations

This is neither a bug nor a feature, maybe just a questionable design decision, but as a user I'd expect that the `postgresl.conf` contains `max_wal_senders = 0` when configuring `config.postgresql.max_wal_senders` to do so. 

(the same applies for `wal_keep_size`)

Relevant part of the Chart: https://github.com/CloudPirates-io/helm-charts/blob/postgres-0.17.2/charts/postgres/templates/configmap.yaml#L90-L101

### Affected Helm charts

postgres

zOnlyKroks:
### Description of the change

Remove condition around wal_level and max_wal_senders

### Benefits

Ability to set wal_level and max_wal_senders even without replication enabled.

### Possible drawbacks

None, backwards compatible

### Applicable issues

- fixes #1160

### Checklist

### DIFF
--- a/charts/postgres/values.yaml
+++ b/charts/postgres/values.yaml
@@ -154,7 +154,7 @@ config:
     # @param config.postgresql.wal_level Determines how much information is written to the WAL - https://postgresqlco.nf/doc/en/param/wal_level/
     wal_level: replica
     # @param config.postgresql.max_wal_senders Specifies the maximum number of concurrent connections from standby servers or streaming base backup clients - https://postgresqlco.nf/doc/en/param/max_wal_senders/
-    max_wal_senders: 5
+    max_wal_senders: 10
     # @param config.postgresql.wal_keep_size Specifies the minimum size of past WAL files kept in the pg_wal directory (in MB) - https://postgresqlco.nf/doc/en/param/wal_keep_size/
     wal_keep_size: 1024
 


btschwertfeger:
Haven't tested it but the change is trivial. What I can see is that the default for `max_wal_senders` is now set to `5` for non-/replication cases due to the default in `values.yaml`. Meaning that this change will affect non-replication setups as Postgres's default value for `max_wal_senders` is `10` (https://www.postgresql.org/docs/17/runtime-config-replication.html#GUC-MAX-WAL-SENDERS). This might not be critical but at least something to note.

BTW: Thanks for the timely response and efforts!

zOnlyKroks:
> Haven't tested it but the change is trivial. What I can see is that the default for `max_wal_senders` is now set to `5` for non-/replication cases due to the default in `values.yaml`. Meaning that this change will affect non-replication setups as Postgres's default value for `max_wal_senders` is `10` (https://www.postgresql.org/docs/17/runtime-config-replication.html#GUC-MAX-WAL-SENDERS). This might not be critical but at least something to note.
> 
> BTW: Thanks for the timely response and efforts!

I shall set it to default 10 then and call it a day

PR #1194, opened 2026-03-27
[all] scoped per-chart changelogs in GitHub Releases

ixxeL2097:
## Summary

Addresses #844 - replace CHANGELOG.md file commits with scoped release notes injected directly into GitHub Releases.

### The problem

The current pipeline generates CHANGELOG.md files per chart and commits them via a bot. This creates several issues:

- **Git history noise**: every chart change triggers a bot commit (`chore: update CHANGELOG.md for merged changes`), doubling the number of commits
- **Unscoped entries**: changelogs reference unrelated charts (e.g., `cluster-operator` versions in the rabbitmq changelog)
- **Hard to assess impact**: users can't tell if an update affects rendered templates or is purely informational
- **Trigger loops**: the bot commit itself can re-trigger the pipeline, requiring author-based skip logic

### The solution

Instead of committing changelog files, generate scoped release notes and inject them into the GitHub Release body created by chart-releaser. This keeps git history clean and puts the changelog where users actually look for it - the release page.

### How it works

```
push to main
  -> detect unreleased charts (version in Chart.yaml has no matching tag)
  -> chart-releaser creates tag + GitHub Release + .tgz package
  -> OCI push + cosign signing
  -> generate-changelog.py release-notes {chart}     <-- NEW
  -> gh release edit {tag} --notes-file ...           <-- NEW
```

The Python script (`scripts/generate-changelog.py`) generates scoped changelogs by:

1. **Path filtering** - `git log -- charts/{chart}/` so only commits touching the chart directory are included
2. **Version grouping** - parses `Chart.yaml` at each commit to detect version boundaries
3. **Noise filtering** - skips automated commits (changelog updates, merges)
4. **Empty bump detection** - marks versions with only `Chart.yaml` changes as "Version bump only"
5. **Conventional commit classification** - breaking changes, features, fixes, image updates, docs, security
6. **Smart author resolution** - GitHub noreply emails to @username, bot accounts to @name[bot], fallback to linked author name

> **Note on commit classification**: the script currently follows [Conventional Commits](https://www.conventionalcommits.org/) patterns (`feat:`, `fix:`, `chore:`, etc.) as it was adapted from one of my personal repos. The classification groups and regex patterns are fully configurable - either via the `DEFAULT_GROUPS` list in the script or through an optional `.changelog-config.yml` file. Feel free to adjust the categories and emojis to match your project's conventions.

### CLI usage

```bash
# Generate release notes for a single chart (for GitHub Release body)
[...]

PR #1249, opened 2026-04-17
fix(ci): use yaml.safe_dump for artifacthub.io/changes annotations

ixxeL2097:
## Summary

Fixes the broken `artifacthub.io/changes` annotations reported by @zOnlyKroks in #1194.

**Root cause**: the inline Python script in the changelog workflow used f-string formatting to build YAML, which did not escape special characters in commit descriptions (`{}`, `[]`, `#`, `>=`, embedded double quotes, etc.). ArtifactHub rejected the annotation with:

```
error preparing package: error enriching package from annotations: 1 error occurred:
  * invalid changes annotation. Please use quotes on strings that include
    any of the following characters: {}:[],&*#?|-<>=!%@
```

**Fix**: replace the manual string building with `yaml.safe_dump()` (PyYAML is already installed in the workflow). This handles quoting and escaping automatically for any commit message content.

### Before (broken on special chars)

```yaml
- kind: fixed
  description: "handle "quoted" values in config: key=val"   # ← broken
- kind: changed
  description: "bump go to >=1.22 #security"                 # ← # parsed as comment
```

### After (safe_dump handles it)

```yaml
- kind: fixed
  description: 'handle "quoted" values in config: key=val'   # ← properly quoted
- kind: changed
  description: 'bump go to >=1.22 #security'                 # ← properly quoted
```

Also passes `REPO_URL` via environment variable instead of shell interpolation inside the Python script for better safety.

## Test plan

- [x] Tested locally with commit messages containing all problematic characters (`{}:[],&*#?|-<>=!%@"`)
- [ ] Verify next chart release generates valid ArtifactHub annotations

### DIFF
--- a/.github/workflows/changelog-and-release.yaml
+++ b/.github/workflows/changelog-and-release.yaml
@@ -196,12 +196,15 @@ jobs:
             echo "Generating ArtifactHub annotation for $CHART_NAME ($COMMIT_RANGE)"
 
             # Generate changes YAML from commit messages
+            # Uses yaml.safe_dump to properly escape special characters
+            # in commit descriptions (fixes ArtifactHub parse errors).
             CHANGES_YAML=$(git log --pretty=format:"%s" "$COMMIT_RANGE" -- "$CHART_DIR/" | \
               grep -v "^Merge " | grep -v "^chore: update CHANGELOG" | \
-              python3 -c "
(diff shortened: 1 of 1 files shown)


zOnlyKroks:
Thank you for your swift fix :)

PR #1281, opened 2026-04-28
[keycloak]: Default tls truststoreExistingSecret

### LINKED ISSUE #1269: Keycloak `tls.truststoreExistingSecret`
Not sure if this is a bug or a feature request . 
This is related to keycloak , latest version and the environment is non-specific (except the certificate :)) )

If I am using an external certificate , and provide the truststore.jks in a secret to be used by keycloak, I also have to provide the secret name from which it can get the jks file. 
https://github.com/CloudPirates-io/helm-charts/blob/d3efd38389171337f61f4e41ca0cc7f6af6a7c90/charts/keycloak/templates/statefulset.yaml#L367

Would it be possible to have a default value for this secret ? 

I thought it was defined here , https://github.com/CloudPirates-io/helm-charts/blob/d3efd38389171337f61f4e41ca0cc7f6af6a7c90/charts/keycloak/templates/_helpers.tpl#L225 , but I don't think this defined value is used in the templates. 

It would benefit me to have a default value for this truststore secret , as I have other conditions if the external certificate is there or not. And this way I could use my condition only on the enable truststore value and the rest would fall in place. 
Can you help ?

zOnlyKroks:
### Description of the change

Default tls truststoreExistingSecret

### Benefits

More thurrough tls by default

### Applicable issues

- fixes #1269

### Checklist

### DIFF
--- a/charts/keycloak/templates/statefulset.yaml
+++ b/charts/keycloak/templates/statefulset.yaml
@@ -161,7 +161,7 @@ spec:
             - name: KC_HTTPS_TRUST_STORE_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ .Values.tls.truststoreExistingSecret | default (include "keycloak.fullname" .) }}
+                  name: {{ include "keycloak.truststoreSecretName" . }}
                   key: {{ .Values.tls.truststoreExistingSecretKey }}
             {{- end }}
             {{- if .Values.database.type }}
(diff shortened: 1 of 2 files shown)


charts/keycloak/templates/statefulset.yaml
```diff
@@ -367,10 +367,10 @@ spec:
         {{- if .Values.tls.truststoreEnabled }}
         - name: tls-truststore
           secret:
-            secretName: {{ .Values.tls.truststoreExistingSecret }}
+            secretName: {{ include "keycloak.truststoreSecretName" . }}
             items:
-              - key: truststore.jks
-                path: truststore.jks
+              - key: {{ .Values.tls.truststoreExistingSecretKey }}
```
kutekahuna, line 372: I know it's late , but can you make this key separate from this line ? https://github.com/CloudPirates-io/helm-charts/pull/1281/changes#diff-e47a2c290d6eee0973b6ce9cbb8b8c31b60128b7f65132595785debf8f74a4f0R164 
these 2 should be 2 different keys as one refers to the truststore file , and the other one refers to the password that truststore is encrypted with.

PR #1351, opened 2026-06-01
[kafka] Add Apache Kafka chart

### LINKED ISSUE #1350: [Feature] Add an Apache Kafka chart
### Name and Version

kafka 0.1.0 (appVersion 4.2.1)

The repository doesn't yet ship an Apache Kafka chart. Kafka is one of the most common building blocks for event-streaming and data-pipeline workloads, so anyone needing it on Kubernetes today has to reach for a heavier third-party chart or roll their own. A first-party chart that follows this repo's conventions (schema validation, tests, Cosign signing) would fill that gap.

I've already built a Kafka chart and would like to propose contributing it. Highlights:

- Deploys Kafka in **KRaft mode** using the official `apache/kafka` image — **no ZooKeeper required**
- Each replica runs as a combined **broker + controller** node, keeping the deployment self-contained and simple to operate for small/medium clusters
- `StatefulSet` with stable pod DNS plus a headless service for inter-node controller traffic
- **SASL authentication** support
- **TLS** support
- Follows the repo's conventions: `values.schema.json`, a Helm test suite, and Cosign signing

I want to be upfront that **this chart was genuinely difficult to get right.** Kafka's configuration surface is large and a few areas took real iteration, wiring **SASL authentication** correctly across brokers, getting **TLS** (certs, listeners, inter-broker comms) working end to end, and reconciling listener / advertised-listener configuration with the templating. It's had serious work put into it, and it would benefit from a careful review.

**Would you like to include this chart in the repo?** If so, I'm happy to open a PR and adjust anything to match your preferences.

- **Bitnami's Kafka chart** — heavier, deprecated, and its image/distribution model differs from this repo's "official upstream image + KRaft" approach.
- **Maintaining the chart out-of-tree** in a personal fork — works, but it doesn't benefit the wider community or get the signing/testing this repo provides.

### Additional informations

Happy to share the chart, walk through the trickier SASL/TLS parts, or split the contribution however makes review easiest.

vaggeliskls:
### Description of the change

Adds a new first-party `kafka` chart (v0.1.0, appVersion 4.2.1) that deploys Apache Kafka in **KRaft mode** using the official `apache/kafka` image (pinned by digest) — **no ZooKeeper required**.

Each replica runs as a combined **broker + controller** node, deployed via a `StatefulSet` with stable pod DNS and a headless Service for inter-node controller traffic. Broker traffic is split across separate `CLIENT` (9092), `INTERNAL` (9094), and `CONTROLLER` (9093) listeners; the internal port is exposed only on the headless Service.

Highlights:

- **SASL/PLAIN authentication** (`auth.*`) on client, inter-broker, and controller listeners — enabled by default with auto-generated passwords stored in a Secret and preserved across upgrades. Supports bring-your-own credentials (`auth.existingSecret`) and per-listener toggles.
- **TLS** (`tls.enabled`) on all listeners — listeners become `SSL`, or `SASL_SSL` when the matching auth boundary is enabled. Three certificate sources (`tls.source`): `self-signed` (chart-generated CA + cert, persisted across upgrades), `existing-secret`, and `cert-manager`. Optional mutual TLS via `tls.clientAuth`. The PEM keystore is assembled at runtime, compatible with a read-only root filesystem.
- **Hardened pod defaults**: non-root, read-only root filesystem, all capabilities dropped, seccomp `RuntimeDefault`, `automountServiceAccountToken: false`. Works with hardened images (e.g. Docker Hardened Images).
- **Observability**: optional kafka-exporter Deployment and Prometheus `ServiceMonitor` (`metrics.*`) that authenticate over SASL and connect over TLS when enabled.
- **Resilience & ops**: quorum-aware PodDisruptionBudget, configurable persistence (PVC templates or existing claim), optional NetworkPolicy, console-only log4j2 logging, a startup probe sized for KRaft quorum formation, and `extraEnvVars` / `extraVolumes` / `extraObjects` escape hatches.

Follows the repo's conventions: `values.schema.json`, a Helm unit-test suite (defaults, auth, TLS, OpenShift, common parameters), `README.md`, `CHANGELOG.md`, and the `common` chart dependency.

### Benefits

Provides a first-party, upstream-image, KRaft-based Kafka chart so users no longer have to reach for a heavier/deprecated third-party chart or roll their own. Ships with the schema validation, testing, and Cosign signing this repo provides.

### Possible drawbacks

- Combined broker + controller topology targets small/medium clusters; dedicated controller/broker roles are not currently supported.
- Only SASL/PLAIN is implemented (not SCRAM/OAuth/mTLS-as-auth).
- The SASL/TLS listener wiring is the most intricate part of the chart and would benefit from careful review.

### Applicable issues

- fixes #1350

### Additional information

The chart was substantial to get right — particularly wiring SASL across broker/inter-broker/controller listeners, end-to-end TLS (certs, listeners, inter-broker comms), and reconciling listener / advertised-listener configuration with the templating. Happy to walk through the trickier parts or adjust anything to match your preferences.

### Checklist


zOnlyKroks:
First of all, thank you for this massive improvement. One last tiny thing before i properly review it, mind adding a artifacthub-repo.yml to the charts base folder? We will provide the proper repositoryId at a later date.
See: https://github.com/CloudPirates-io/helm-charts/blob/main/charts/postgres/artifacthub-repo.yml
Thank you

dloewen2:
Hi @vaggeliskls, 
thanks for this contribution, i will also test around with your PR. 
For now i created the dockerhub and artifacthub-repo. Could you please add this id in the artifacthub-repo.yaml: 
0c7f8fd8-4d25-4be6-8147-8f702874a574

charts/kafka/README.md
```diff
@@ -0,0 +1,391 @@
...
+  storageClass: fast-ssd
+resources:
+  requests:
+    cpu: 500m
+    memory: 2Gi
+  limits:
+    memory: 4Gi
+kafkaConfig:
+  defaultReplicationFactor: 3
+  minInsyncReplicas: 2
```
dloewen2: This value is not being used in any other file.

charts/kafka/values.yaml
```diff
@@ -0,0 +1,391 @@
...
+  periodSeconds: 10
+  ## @param startupProbe.timeoutSeconds Timeout seconds for startupProbe
+  timeoutSeconds: 5
+  ## @param startupProbe.failureThreshold Failure threshold for startupProbe
+  failureThreshold: 30
+  ## @param startupProbe.successThreshold Success threshold for startupProbe
+  successThreshold: 1
+
+## @param command Override default container command (useful when the default entrypoint needs to be replaced)
+command: []
```
dloewen2, line 372: The added command seems be added wrong. When i set the command to :
```
command:
- sleep 3600
```
i get this last line in the args:
```
/opt/kafka/bin/kafka-storage.sh format -t "${KAFKA_CLUSTER_ID}" -c "$RUNTIME_CONFIG" --ignore-formatted"sleep 3600"
```
I think the sleep command should be added in a new line. 

---

Apart from this, is it correct, that the command value only overrides the execute command in the script given in the args? My first expectation was, that it completely overrides the command attribute of the kafka statefulset.

charts/kafka/values.yaml
```diff
@@ -0,0 +1,391 @@
...
+
```
dloewen2, line 377: If we can override the command attribute, then an override for the args would also be good to have. 
Also an 'extraArgs' value would seem to be a good addition.

charts/kafka/README.md
```diff
@@ -0,0 +1,391 @@
...
+  extraConfig:
+    - "min.insync.replicas=2"
+```
+
```
dloewen2, line 404: Would an example for a deployment with 1 replica be useful? 
Something like a dev-deployment with very little resources. (I am no expert with kafka, so i don't know if a deployment with 1 replica is something that makes sense. Or if you just edit the replicas to 1 and then its done?)

charts/kafka/templates/metrics-deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+{{- . | indent 4 }}
+  {{- end }}
+spec:
+  replicas: 1
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  selector:
+    matchLabels:
+      {{- include "kafka.metrics.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
```
dloewen2, line 20: Please add the option to add annotations for the metrics pod-template. 
With this you can add the prometheus-scrape annotations to the pod, so the metrics can be used by prometheus without the servicemonitor.

charts/kafka/Chart.yaml
```diff
@@ -0,0 +1,45 @@
...
+    email: hello@cloudpirates.io
+    url: https://www.cloudpirates.io
+  - name: vaggeliskls
+    email: vaggelis.vk@gmail.com
+    url: https://github.com/vaggeliskls
+dependencies:
+  - name: common
+    version: "2.2.0"
+    repository: oci://registry-1.docker.io/cloudpirates
+icon: https://kafka.apache.org/images/apache-kafka.png
```
dloewen2: Please use this icon:
https://a.storyblok.com/f/143071/512x512/a5d4be1ea1/kafka-logo.svg

charts/kafka/README.md
vaggeliskls: Good catch, that line was a leftover and did nothing (kafkaConfig.minInsyncReplicas isn't a real value). Removed it. The extraConfig: ["min.insync.replicas=2"] line right below is the correct way to set it.

charts/kafka/values.yaml
vaggeliskls: You're completely right, thank you. It was a Go-template whitespace bug the {{- if }} was trimming the newline.

charts/kafka/values.yaml
vaggeliskls: Added extraArgs it appends to kafka-server-start.sh (handy for --override key=value), and is ignored when command is set.

charts/kafka/templates/metrics-deployment.yaml
vaggeliskls: Added metrics.podAnnotations

charts/kafka/Chart.yaml
vaggeliskls: Done

charts/kafka/README.md
vaggeliskls: You were right to be unsure, with the old defaults a single-node deploy came up looking healthy but broke as soon as a consumer group or transaction tried to create its internal topic at replication factor 3.

So I added a single-node example, and changed how the replication factors work. They're now automatically capped to replicaCount, since a replication factor can never exceed the number of brokers. So replicaCount: 1 just works, replicaCount: 3 renders exactly as before, and explicit values are still respected (just capped if impossible). I also added a "Cluster sizing" section covering the nodes/replicas/replication relationship and the quorum math. One caveat I documented: Kafka won't re-replicate already-created internal topics when you scale up later, which is a Kafka limitation rather than something the chart can fix.

I also tested it on a single-node k3d cluster: produce/consume works and __consumer_offsets gets created at replication factor 1 automatically, which is the exact thing that used to fail with the old defaults.

charts/kafka/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "kafka.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+    - Egress
+  egress:
+    - {}
+  ingress:
```
zOnlyKroks, line 22: templates/networkpolicy.yaml:22-32 opens ingress only for the client (9092) and controller (9093) ports. The INTERNAL / inter-broker listener (9094) has no ingress rule, so it's denied.

Since inter.broker.listener.name=INTERNAL (configmap.yaml:18) and brokers advertise INTERNAL://<pod>:9094, all broker-to-broker data-plane traffic (replication, fetch) flows over 9094. With networkPolicy.enabled=true and replicaCount > 1 (both the realistic case, default is 3), replication breaks and the cluster won't reach ISR. The comment on the 9093 rule ("internal controller communication") appears to conflate the INTERNAL listener with the CONTROLLER listener.

zOnlyKroks (changes requested):

charts/kafka/templates/networkpolicy.yaml
vaggeliskls: Good catch, you're right.

zOnlyKroks:
Ill see to it, that I review and merge this by end of week if nothing comes up.

dloewen2:
LGTM, @zOnlyKroks you can merge when you are done with your review.

PR #1390, opened 2026-06-16
[zookeeper] Make secureClientPort optional

### LINKED ISSUE #1369: [zookeeper] Make secureClientPort optional in zoo.cfg
Hello, thanks for maintaining this repository!

I ran across this issue with the recent modifications around secureClientPort for the zookeeper chart. The chart always writes secureClientPort=2281 to zoo.cfg (via templates/configmap.yaml), even when TLS is not configured, which is my case, since I don't need it. 

Once the port setting is present, ZooKeeper expects SSL keystore/truststore config and the pod fails to start with this message:

```shell
java.lang.UnsupportedOperationException: SSL isn't supported in NIOServerCnxn
	at org.apache.zookeeper.server.NIOServerCnxnFactory.configure(NIOServerCnxnFactory.java:621)
	at org.apache.zookeeper.server.quorum.QuorumPeerMain.runFromConfig(QuorumPeerMain.java:174)
	at org.apache.zookeeper.server.quorum.QuorumPeerMain.initializeAndRun(QuorumPeerMain.java:137)
	at org.apache.zookeeper.server.quorum.QuorumPeerMain.main(QuorumPeerMain.java:91)
2026-06-05 14:22:28,293 [myid:0] - INFO  [main:o.a.z.a.ZKAuditProvider@42] - ZooKeeper audit is disabled.
2026-06-05 14:22:28,296 [myid:0] - ERROR [main:o.a.z.u.ServiceUtils@48] - Exiting JVM with code 1
``` 

This is the PR that introduced the change: https://github.com/CloudPirates-io/helm-charts/pull/1212/changes

The | default 2281 in the template also prevents disabling it via values (null still renders as 2281).

Would you be open to a modification where this is an optional configuration? like you did for the admin server block? Let me know what you think and I can open a PR.

Proposed change:
Only emit secureClientPort when explicitly configured (e.g. service.ports.secureClient is set). Default to disabled (null). This follows the same pattern already used for the optional admin server block in configmap.yaml.

Changing the default from implicit 2281 to disabled could be a breaking change: upgrades with no values override would stop exposing the TLS client port. Users who depend on TLS must explicitly set service.ports.secureClient.

Thank you

LuluGO:
### Description of the change

The chart currently always writes `secureClientPort=2281` to `zoo.cfg` and exposes port 2281 on the StatefulSet, Services, and NetworkPolicy, even when TLS is not configured. Once `secureClientPort` is set, ZooKeeper expects SSL keystore configuration and the pod fails to start without it. More info in this issue: https://github.com/CloudPirates-io/helm-charts/issues/1369

This change makes the TLS client port opt-in:
- Default `service.ports.secureClient` to `null` (disabled)
- Only emit `secureClientPort` in `zoo.cfg` when `service.ports.secureClient` is set
- Conditionally expose the `secure-client` port in the StatefulSet, Services, and NetworkPolicy to match

### Benefits
- Plain TCP deployments (port 2181 only) work out of the box without TLS configuration
- Follows the same optional pattern already used for the admin server block in `configmap.yaml`
- Users who need TLS can explicitly set `service.ports.secureClient: 2281` and configure SSL via `zookeeperConfig.extraConfigs` + volume mounts

### Possible drawbacks
**Breaking change:** Upgrades with default values will no longer set `secureClientPort` in `zoo.cfg` or expose port 2281 in Kubernetes. Deployments that relied on the implicit default (without setting `service.ports.secureClient` in values) but had TLS configured via `extraConfigs` will need to explicitly set:
```yaml
service:
  ports:
    secureClient: 2281
```

Setting the port without SSL configuration in zoo.cfg will still cause startup failure, this is expected ZooKeeper behavior, not introduced by this change.

### Applicable issues

- fixes #1369 

### Additional information

Tested locally:

helm dependency update charts/zookeeper
helm unittest charts/zookeeper   # 33/33 passed

```
### Chart [ zookeeper ] charts/zookeeper

 PASS  test zookeeper common parameters charts/zookeeper/tests/common-parameters_test.yaml
 PASS  test openshift related settings  charts/zookeeper/tests/openshift_test.yaml
 PASS  test zookeeper specific parameters       charts/zookeeper/tests/unittest-defaults_test.yaml

Charts:      1 passed, 1 total
Test Suites: 3 passed, 3 total
Tests:       33 passed, 33 total
Snapshot:    0 passed, 0 total
Time:        180.711416ms
```

**Chart version (`0.11.0`):** Bumped as a **minor** release (from `0.10.3`) rather than a patch because the default behavior changes: `secureClientPort` and the `secure-client` K8s port are no longer enabled by default. Per [semver](http://semver.org/), that is not fully backwards compatible for users who relied on the implicit `2281` default without setting `service.ports.secureClient` in values. A patch bump would imply no behavior change for existing installs with default values.

### Checklist

### DIFF
--- a/charts/zookeeper/Chart.yaml
+++ b/charts/zookeeper/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: zookeeper
 description: Apache ZooKeeper is a centralized service for maintaining configuration information, naming, providing distributed synchronization, and providing group services.
 type: application
-version: 0.11.0
+version: 0.12.0
 appVersion: "3.9.5"
 keywords:
   - zookeeper
--- a/charts/zookeeper/README.md
+++ b/charts/zookeeper/README.md
@@ -150,7 +150,7 @@ zkCli.sh -server my-zookeeper:2181
 | ------------------------------ | -------------------------------------------- | ----------- |
 | `service.type`                 | Kubernetes service type                      | `ClusterIP` |
 | `service.ports.client`         | ZooKeeper client service port                | `2181`      |
-| `service.ports.secureClient`   | ZooKeeper secure client service port         | `2281`      |
+| `service.ports.secureClient`   | TLS client port. Set to null to disable (default). When set, also configure SSL via zookeeperConfig.extraConfigs and mount keystore files with extraVolumes/extraVolumeMounts. Setting the port without SSL config will cause startup failure.     | `null`      |
 | `service.ports.quorum`         | ZooKeeper quorum service port                | `2888`      |
 | `service.ports.leaderElection` | ZooKeeper leader election service port       | `3888`      |
 | `service.ports.admin`          | ZooKeeper admin service port                 | `8080`      |
--- a/charts/zookeeper/templates/configmap.yaml
+++ b/charts/zookeeper/templates/configmap.yaml
@@ -16,7 +16,9 @@ data:
     initLimit={{ .Values.zookeeperConfig.initLimit | default 10 }}
     syncLimit={{ .Values.zookeeperConfig.syncLimit | default 5 }}
     clientPort={{ .Values.service.ports.client | default 2181 }}
-    secureClientPort={{ .Values.service.ports.secureClient | default 2281 }}
+    {{- if .Values.service.ports.secureClient }}
+    secureClientPort={{ .Values.service.ports.secureClient }}
+    {{- end }}
     electionPortBindRetry={{ .Values.zookeeperConfig.electionPortBindRetry | default 10 }}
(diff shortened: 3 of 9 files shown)


zOnlyKroks:
This would now need a version bump @LuluGO

LuluGO:
> This would now need a version bump @LuluGO

Rebased on main after external-access merge, and bumped chart version to 0.12.0, thanks!

mms-gianni (approved):
Ran tests and created an instance. 

The default with `.Values.service.ports.secureClient= null` was successful. Using `.Values.service.ports.secureClient= 2281` still crashes as mentioned in the PR. 

We should mention that and probably add a installation instruction on how to configure it.

I've prepared it and will add on a separate PR after some more testing.

PR #1400, opened 2026-06-25
[postgres] Use shared helper for metrics container securityContext

### LINKED ISSUE #1399: [postgres] metrics sidecar hardcodes runAsUser 65534, breaks OpenShift restricted SCC
## Summary

When `metrics.enabled: true`, the postgres-exporter sidecar in the postgres chart hardcodes `runAsUser: 65534` / `runAsGroup: 65534`. On OpenShift clusters using restricted SCC, this prevents the **entire** StatefulSet pod from being created, because the namespace-assigned UID range does not include `65534`.

The main postgres container is already OpenShift-compatible via `cloudpirates.renderContainerSecurityContext` and `targetPlatform: openshift`, but the metrics sidecar bypasses that logic.

## Symptoms

StatefulSet stuck at `0/1` ready with events like:

```
create Pod <name>-0 failed: pods "<name>-0" is forbidden: unable to validate against any security context constraint:
...
provider restricted-v2: .containers[1].runAsUser: Invalid value: 65534: must be in the ranges: [1004460000, 1004469999]
```

`containers[1]` is the `metrics` sidecar (`quay.io/prometheuscommunity/postgres-exporter`).

## Root cause

In `charts/postgres/templates/statefulset.yaml`, the postgres container correctly uses the shared helper:

```yaml
securityContext: {{ include "cloudpirates.renderContainerSecurityContext" . | nindent 12 }}
```

The metrics container does not. It inlines a fixed security context:

```yaml
securityContext:
  ...
  runAsUser: 65534
  runAsGroup: 65534
```

This was introduced when metrics support was added in #285. OpenShift support for the main postgres container landed earlier in #184, using the common helper from #166 — but the metrics sidecar was not updated to follow the same pattern.

### redis

The redis chart metrics container already uses the shared helper:

```yaml
- name: metrics
  securityContext: {{ include "cloudpirates.renderContainerSecurityContext" . | nindent 12 }}
```

With `targetPlatform: openshift`, the helper omits `runAsUser` / `runAsGroup` so OpenShift assigns a UID from the namespace range.

### valkey

Valkey had the same hardcoded `65534` problem and was fixed in #1355 (merged 2026-06-03):

> The metrics container previously hardcoded `runAsUser: 65534` / `runAsGroup: 65534` (nobody). It now inherits the chart's standard container security context via `cloudpirates.renderContainerSecurityContext`.

### mongodb

MongoDB exposes `metrics.containerSecurityContext` in values, allowing consumers to override (e.g. `runAsUser: null` on OpenShift). Postgres does not expose an equivalent override today.

## Why this is surfacing now

Metrics were disabled by default when added (#285). Downstream deployments on OpenShift ran fine with `targetPlatform: openshift` because only the main postgres container was rendered.

Recently, monitoring was enabled across several production gitops environments by setting:

```yaml
metrics:
  enabled: true
```

on postgres subcharts bundled in application charts (e.g. endorser-service). That rolled out the metrics sidecar for the first time on OpenShift. Both wallet and controller postgres StatefulSets in the same namespace failed simultaneously, taking down dependent services even though PVC data was intact.

Disabling metrics (`metrics.enabled: false`) is the only practical workaround until the chart is fixed, because there is no `metrics.containerSecurityContext` values knob like mongodb provides.

## Suggested fix

Apply the same change as #1355 (valkey) to postgres:

```yaml
- name: metrics
  securityContext: {{ include "cloudpirates.renderContainerSecurityContext" . | nindent 12 }}
```

Optionally also expose `metrics.containerSecurityContext` (mongodb pattern) for finer control, but the shared helper + `targetPlatform: openshift` should be sufficient for restricted SCC clusters.

## Affected versions

Confirmed on chart `postgres-0.15.5` (postgres 18.1.0). Still present on `main` as of postgres chart **0.19.5** — the metrics template still hardcodes `65534`.

## Workaround

```yaml
metrics:
  enabled: false
```

## Environment

- Platform: OpenShift 4.x (restricted-v2 SCC)
- Chart dependency: `postgres` subchart via application charts (e.g. openwallet-foundation/endorser-service)
- `targetPlatform: openshift` set on postgres values (main container works; metrics sidecar does not)

PatStLouis:
### Description of the change

The `metrics` (postgres-exporter) container in the postgres StatefulSet defined its `securityContext` as a hardcoded inline block with `runAsUser: 65534` / `runAsGroup: 65534`, while the postgres container renders through the shared `cloudpirates.renderContainerSecurityContext` helper. This change switches the metrics container to use the same helper.

This is the same convention already followed by the `redis` chart and was applied to `valkey` in #1355.

### Benefits

- Consistency: a single source of truth for container security context across all containers in the chart.
- The metrics container now honors `targetPlatform: openshift` and omits fixed UIDs so OpenShift restricted SCC can assign namespace UIDs.
- Fixes pod creation failures when `metrics.enabled: true` on OpenShift clusters.

### Possible drawbacks

The metrics container previously hardcoded `runAsUser: 65534` / `runAsGroup: 65534` (nobody). It now inherits the chart's standard container security context (`runAsUser: 999` / `runAsGroup: 999` by default on vanilla Kubernetes). The postgres-exporter image runs fine as non-root, but anyone relying on the metrics container running specifically as `65534` should set `.Values.containerSecurityContext` accordingly.

### Applicable issues

Fixes #1399

### Additional information

Verified with:

```bash
helm template test . --set metrics.enabled=true --set targetPlatform=openshift
```

The metrics container renders without `runAsUser` / `runAsGroup` on OpenShift.

Added a unit test in `tests/openshift_test.yaml` for metrics + `targetPlatform: openshift`.

### Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/charts/postgres/CHANGELOG.md
+++ b/charts/postgres/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this chart will be documented in this file.
 
+## [0.19.6] - 2026-06-25
+
+- [postgres] Use shared helper for metrics container securityContext on OpenShift (#1399)
+
 ## [0.18.3] - 2026-03-18
 
 - Update image.repository to a9abf42 (#1164) ([3d4a6b9b](https://github.com/CloudPirates-io/helm-charts/commit/3d4a6b9b))
--- a/charts/postgres/Chart.yaml
+++ b/charts/postgres/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: postgres
 description: The World's Most Advanced Open Source Relational Database
 type: application
-version: 0.19.5
+version: 0.19.6
 appVersion: "18.4.0"
 keywords:
   - postgres
@@ -42,10 +42,10 @@ annotations:
     - name: Maintainer CloudPirates
       url: https://www.cloudpirates.io
   artifacthub.io/changes: |-
-    - kind: added
-      description: "Add namespaceOverride Value to all Charts that missing it"
+    - kind: fixed
+      description: "Use shared security context helper for metrics sidecar on OpenShift"
       links:
-        - name: "Pull Request"
-          url: "https://github.com/CloudPirates-io/helm-charts/pull/1214"
+        - name: "Issue"
+          url: "https://github.com/CloudPirates-io/helm-charts/issues/1399"
         - name: "Changelog"
           url: "https://github.com/CloudPirates-io/helm-charts/blob/main/charts/postgres/CHANGELOG.md"
--- a/charts/postgres/templates/statefulset.yaml
+++ b/charts/postgres/templates/statefulset.yaml
@@ -234,16 +234,7 @@ spec:
         - name: metrics
           image: "{{ .Values.metrics.image.registry }}/{{ .Values.metrics.image.repository }}:{{ .Values.metrics.image.tag }}"
           imagePullPolicy: {{ .Values.metrics.image.pullPolicy }}
-          securityContext:
-            allowPrivilegeEscalation: false
-            capabilities:
-              drop:
-                - ALL
-            privileged: false
-            readOnlyRootFilesystem: true
-            runAsNonRoot: true
-            runAsUser: 65534
-            runAsGroup: 65534
+          securityContext: {{ include "cloudpirates.renderContainerSecurityContext" . | nindent 12 }}
           env:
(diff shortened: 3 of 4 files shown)


PatStLouis:
Superseded by #1401 (same change from OpSecId org fork).

PR #1212, opened 2026-04-08
[zookeeper] feat: enable TLS and authenticated clients

tdannenmuller:
### Description of the change

This PR enables ZooKeeper TLS and authenticated client connectivity by adding a dedicated secure client port and exposing related configuration paths needed to wire secure client traffic end to end.

The change includes:
- Secure client service port support in chart values and schema
- Template updates so the secure client port is exposed consistently across services, pod ports, and network policy rules
- ZooKeeper configuration support for secure client port settings
- Documentation and tests updates aligned with the new configuration surface
- Chart version bump for this feature release

### Benefits

- Enables secure client traffic path configuration for ZooKeeper deployments
- Improves support for TLS and authenticated client scenarios
- Keeps chart docs, schema validation, and test coverage aligned with new values

### Possible drawbacks

- Additional port exposure and config surface slightly increases deployment complexity
- Secure client port support alone does not fully enforce TLS and authentication unless users also provide the required ZooKeeper security settings and certificates in their deployment

### Applicable issues

- fixes #N/A

### Additional information

Validation performed:
- Helm lint passed for the chart
- End to end chart deployment with clients apps connected over tls and sasl

### Checklist

- [X] Chart version bumped in [charts/zookeeper/Chart.yaml](charts/zookeeper/Chart.yaml) according to [semver](http://semver.org/). This is not necessary when the changes only affect README.md files.
- [X] Variables are documented in [charts/zookeeper/values.yaml](charts/zookeeper/values.yaml) and added to [charts/zookeeper/README.md](charts/zookeeper/README.md)
- [X] Title of the pull request follows this pattern [<name_of_the_chart>] Descriptive title
- [X] All commits are signed and verified (required as of January 22, 2026 - see [CONTRIBUTING.md](CONTRIBUTING.md#setting-up-your-development-environment))

PR #1409, opened 2026-07-03
[zookeeper] Fix startup crash by gating secure client port behind tls.enabled

### LINKED ISSUE #1369: [zookeeper] Make secureClientPort optional in zoo.cfg

mms-gianni:
### Description of the change

Since chart version 0.9.0 (#1212), secureClientPort is rendered unconditionally into zoo.cfg. ZooKeeper only supports a secure client port with the Netty connection factory, which the chart never configures, so every install with default values crashes on startup with:

```
java.lang.UnsupportedOperationException: SSL isn't supported in NIOServerCnxn
	at org.apache.zookeeper.server.NIOServerCnxnFactory.configure(NIOServerCnxnFactory.java:621)
```
There is also no values-level escape hatch, since
`secureClientPort={{ .Values.service.ports.secureClient | default 2281 }}` re-applies the
default even when the value is nulled out.

This PR makes the secure client port opt-in via a new `tls.enabled` value (default `false`,
matching the convention used by other charts in this repo):

- `zoo.cfg` only contains `secureClientPort` when `tls.enabled=true`, and then also sets
  `serverCnxnFactory=org.apache.zookeeper.server.NettyServerCnxnFactory`, which TLS requires
- The `secure-client` port on the client service, headless service, StatefulSet container,
  and NetworkPolicy ingress rule is now rendered only when `tls.enabled=true`
- Certificate configuration (keystore/truststore) is supplied via
  `zookeeperConfig.extraConfigs` and `extraVolumes`/`extraVolumeMounts`, documented with an
  example in the README
- Unit tests cover both the default (no secure port rendered) and the `tls.enabled=true` case

### Benefits

- The chart works again with default values — currently every fresh install of
  chart >= 0.9.0 crashloops
- TLS remains fully usable: enabling `tls.enabled` now actually works, because the Netty
  connection factory is set alongside the secure port (previously TLS via `extraConfigs`
  alone would still crash on the NIO factory)

### Possible drawbacks

- Deployments that set `serverCnxnFactory=org.apache.zookeeper.server.NettyServerCnxnFactory`
  via `zookeeperConfig.extraConfigs` as a workaround must now set `tls.enabled=true` to keep
  the secure port exposed. Setting both keeps working (the duplicated `serverCnxnFactory`
  line is harmless but can be removed from `extraConfigs`).
- Services and NetworkPolicies no longer expose port 2281 by default. Since the pod never
  started with the previous defaults, no working deployment can depend on it.

### Applicable issues

- fixes https://github.com/CloudPirates-io/helm-charts/issues/1369

### Additional information

The original intent of #1212 is preserved: `secureClientPort=2281` was documented as an
opt-in `extraConfigs` example in `values.yaml` before it was hard-coded into the ConfigMap.
Chart version is bumped to 0.12.0 (minor, new `tls.enabled` value plus changed default
rendering).
### Checklist

### DIFF
--- a/charts/zookeeper/Chart.yaml
+++ b/charts/zookeeper/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: zookeeper
 description: Apache ZooKeeper is a centralized service for maintaining configuration information, naming, providing distributed synchronization, and providing group services.
 type: application
-version: 0.11.0
+version: 0.12.0
 appVersion: "3.9.5"
 keywords:
   - zookeeper
--- a/charts/zookeeper/README.md
+++ b/charts/zookeeper/README.md
@@ -124,6 +124,25 @@ zkCli.sh -server my-zookeeper:2181
 | `zookeeperConfig.admin.commandUrl`          | Admin server command URL                            | `/commands` |
 | `zookeeperConfig.extraConfigs`              | Extra ZooKeeper configuration lines appended to zoo.cfg | `[]`    |
 
+### TLS Configuration
+
+| Parameter     | Description                                                | Default |
+| ------------- | ---------------------------------------------------------- | ------- |
+| `tls.enabled` | Expose the ZooKeeper secure client port for TLS connections | `false` |
+
+Enabling TLS opens the secure client port (`service.ports.secureClient`) and switches ZooKeeper to the Netty connection factory, which is required for TLS. Supply the certificate configuration via `zookeeperConfig.extraConfigs` and mount the keystore files with `extraVolumes` / `extraVolumeMounts`, for example:
+
+```yaml
+tls:
+  enabled: true
+zookeeperConfig:
+  extraConfigs:
+    - "ssl.keyStore.location=/certs/keystore.jks"
+    - "ssl.keyStore.password=changeit"
+    - "ssl.trustStore.location=/certs/truststore.jks"
+    - "ssl.trustStore.password=changeit"
+```
+
 ### Metrics
 
 | Parameter                                  | Description                                                                      | Default     |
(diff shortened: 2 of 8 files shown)


mms-gianni:
Just saw ... this was already fixed by @LuluGO in #1390

PR #1421, opened 2026-07-09
[valkey] Fix broken Sentinel preStop failover hook on Alpine image

### LINKED ISSUE #1411: valkey chart: preStop graceful-failover hook is broken on the default Alpine image (two bugs)
# valkey chart: preStop graceful-failover hook is broken on the default Alpine image (two bugs)

**Chart:** `valkey` (tested `0.23.0`, also present in `0.20.2`)
**Architecture:** `replication` + `sentinel.enabled: true`

## Summary

The StatefulSet's `preStop` hook is supposed to perform a graceful master
handoff on shutdown (`CLIENT PAUSE` → `SENTINEL FAILOVER` → wait). In practice it
never runs successfully, so a planned master pod termination (e.g. a rolling
update from an image bump) drops the master ungracefully and clients stall until
Sentinel reactively detects the dead master. There are **two independent bugs**.

### Relation to #781 (not a duplicate)

#781 (`[Bug] Downtime occurs during helm upgrade with Sentinel enabled`) fixed a
*different* root cause **for the redis chart** — the Sentinel container had no
preStop — by adding `sentinel.preStop.enabled`,
`sentinel.redisShutdownWaitFailover` and a `terminationGracePeriodSeconds` value.
The **valkey** chart (latest: 0.23.0) never received any of these knobs (it
exposes none of them), still has a preStop on the Valkey container only, and
additionally suffers from the two bugs below, which #781 does not touch. #459 was
about initial HA setup and is unrelated.

---

## Bug 1 — `preStop` uses `/bin/bash`, but the default image has no bash

`templates/statefulset.yaml` hardcodes:

```yaml
lifecycle:
  preStop:
    exec:
      command:
        - /bin/bash
        - /scripts/prestop.sh
```

The chart's default image (`valkey/valkey:*-alpine`) is Alpine-based and ships
**no `/bin/bash`** — only `/bin/sh` (ash). This is true for the official
`valkey-io/valkey-container` Alpine images across versions (`9.0/alpine`,
`9.1/alpine`): the runtime stage runs `apk add --no-cache tzdata setpriv openssl
libgcc` and never installs bash.

**Result:** kubelet cannot exec the hook →

```
FailedPreStopHook: Exec lifecycle hook ([/bin/bash /scripts/prestop.sh]) ...
exec: "/bin/bash": stat /bin/bash: no such file or directory
```

The graceful failover never runs. On a rolling update the master (e.g.
`valkey-0`, updated last) is SIGTERM/SIGKILLed without handoff, and write clients
stall for the Sentinel `down-after-milliseconds` window + promotion.

Note that `data.prestop.sh` in `templates/prestop-configmap.yaml` uses bash-only
constructs (arrays `arr=(...)` / `${arr[@]}`, `[[ ]]`, `((attempt++))`), so
simply switching the shell to `/bin/sh` is **not** sufficient — the script must
be made POSIX-sh compatible.

## Bug 2 — the hook talks to Sentinel via a Service that doesn't expose the Sentinel port

In `templates/prestop-configmap.yaml`:

```sh
VALKEY_SERVICE="{{ include "valkey.fullname" . }}.{{ include "cloudpirates.namespace" . }}.svc.{{ .Values.clusterDomain }}"
...
run_sentinel_command() {
    valkey-cli -h "$VALKEY_SERVICE" -p "$SENTINEL_PORT" ... sentinel "$@"
}
```

`run_sentinel_command` targets the **main** Service (`{{ valkey.fullname }}`) on
the Sentinel port (26379). But `templates/service.yaml` only exposes the Valkey
port on that Service — the Sentinel port (26379) is exposed **only** on the
`-headless` Service and the `-sentinel` Service:

```yaml
# service.yaml (main ClusterIP service) — only the valkey port:
ports:
  - port: {{ .Values.service.port }}
    name: valkey
```

So even after Bug 1 is fixed, `SENTINEL FAILOVER` and
`sentinel get-master-addr-by-name` in the preStop connect to a Service with no
`26379` endpoint and fail. Worse: `sentinel_failover_finished()` treats an empty
reply as "failover complete" (`return 0`), so the hook silently believes the
handoff succeeded when it never happened.

## Impact

Graceful master handoff on planned shutdown does not work out of the box for the
default (Alpine) install with `sentinel.enabled: true`. Every master-pod
rollout produces a `FailedPreStopHook` event and an avoidable client-facing write
outage.

## Suggested fixes

1. Rewrite `prestop.sh` to be POSIX-sh compatible (drop arrays / `[[ ]]` /
   `(( ))`) and invoke it with `/bin/sh` — or `apk add bash` in the Alpine image
   / document that a bash-containing image is required.
2. Point the Sentinel calls in `prestop.sh` at an endpoint that actually serves
   26379. Simplest and most robust: talk to the **local** Sentinel sidecar in the
   same pod at `127.0.0.1:{{ .Values.sentinel.port }}` (it always exists when
   `sentinel.enabled`), rather than the main Service. Alternatively use the
   `-sentinel` / `-headless` Service name.
3. Consider porting the #781 solution to the valkey chart for parity: a preStop on
   the Sentinel container too, plus configurable `terminationGracePeriodSeconds`
   and `sentinel.preStop.enabled` / `redisShutdownWaitFailover` equivalents.

## Environment

- Chart `valkey` 0.23.0 (bug also in 0.20.2)
- Image `valkey/valkey:9.1.0-alpine3.23`
- `architecture: replication`, `sentinel.enabled: true`, 3 replicas
- Kubernetes (any) — reproduces on a plain rolling restart of the StatefulSet

TheDadi:
### Description of the change

The Sentinel `preStop` hook in the valkey chart is supposed to gracefully hand
off the master on shutdown (`CLIENT PAUSE` → `SENTINEL FAILOVER` → wait), but on
the default Alpine image it never actually runs. There are two separate problems
(both reported in #1411):

First, the hook is invoked with `/bin/bash`, and the default
`valkey/valkey:*-alpine` image doesn't have bash, only `/bin/sh` (BusyBox ash).
So kubelet just fails the hook with `exec: "/bin/bash": no such file or
directory`. On top of that, `prestop.sh` itself was written in bash (arrays,
`[[ ]]`, `(( ))`), so pointing it at `/bin/sh` wasn't enough on its own, the
script is rewritten to be POSIX sh.

Second, even once the script runs, the Sentinel calls go to the main Service on
port 26379, but that Service only exposes the Valkey port. 26379 only lives on
the `-headless` and `-sentinel` Services, so `SENTINEL FAILOVER` and
`get-master-addr-by-name` were talking to nothing. Those calls now go to the
local Sentinel sidecar (`127.0.0.1`/`::1`), which is always in the same pod and
actually serves 26379.

Also bumps the chart `0.23.0` → `0.23.1`.

### Benefits

Graceful master handoff on a planned shutdown finally works on a default Alpine +
Sentinel install. No more `FailedPreStopHook` events on every master rollout, and
no more write outage while Sentinel waits to notice the master is gone on its own.

### Possible drawbacks

Nothing for a default install, the hook and configmap still only render with
`architecture: replication` and `sentinel.enabled: true`, and no defaults change.

### Applicable issues

- fixes #1411

### Additional information

Checked a few ways before opening: `helm template`/`helm lint` are clean and the
rendered command is `/bin/sh /scripts/prestop.sh`; the rendered `prestop.sh`
passes both `sh -n` and strict-POSIX `dash -n`; and the whole script was run under
`dash` with a stubbed `valkey-cli` to confirm the
detect-master → pause → failover → wait-for-new-master flow runs end to end.

### Checklist

### DIFF
--- a/charts/valkey/Chart.yaml
+++ b/charts/valkey/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: valkey
 description: High performance in-memory data structure store, fork of Redis. Valkey is an open-source, high-performance key/value datastore that supports a variety of workloads such as caching, message queues, and can act as a primary database.
 type: application
-version: 0.23.0
+version: 0.23.1
 appVersion: "9.1.0"
 home: https://www.valkey.io
 sources:
--- a/charts/valkey/templates/prestop-configmap.yaml
+++ b/charts/valkey/templates/prestop-configmap.yaml
@@ -12,7 +12,7 @@ metadata:
   {{- end }}
 data:
   prestop.sh: |
-    #!/bin/bash
+    #!/bin/sh
 
     # Valkey master failover script for graceful shutdown
     # Based on Redis/Bitnami charts failover handling
@@ -24,7 +24,6 @@ data:
     SENTINEL_PORT="{{ .Values.sentinel.port }}"
     MASTER_NAME="{{ .Values.sentinel.masterName }}"
     HEADLESS_SERVICE="{{ include "valkey.fullname" . }}-headless.{{ include "cloudpirates.namespace" . }}.svc.{{ .Values.clusterDomain }}"
-    VALKEY_SERVICE="{{ include "valkey.fullname" . }}.{{ include "cloudpirates.namespace" . }}.svc.{{ .Values.clusterDomain }}"
 
     # Set authentication if enabled
     {{- if .Values.auth.enabled }}
@@ -40,14 +39,13 @@ data:
 
     # Function to run Valkey commands
     run_valkey_command() {
-        local args=("-h" "$VALKEY_LOOPBACK" "-p" "$VALKEY_PORT")
-        valkey-cli "${args[@]}" "$@"
+        valkey-cli -h "$VALKEY_LOOPBACK" -p "$VALKEY_PORT" "$@"
     }
 
     # Function to check if current instance is master
     is_master() {
         VALKEY_ROLE=$(run_valkey_command role | head -1)
-        [[ "$VALKEY_ROLE" == "master" ]]
+        [ "$VALKEY_ROLE" = "master" ]
     }
 
     # Function to get full hostname for a pod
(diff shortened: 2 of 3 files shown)


zOnlyKroks:
Good Morning @TheDadi ,
mind fixing the merge conflict?
Kind regards

TheDadi:
@zOnlyKroks these changes were merged with https://github.com/CloudPirates-io/helm-charts/pull/1422 therefore i close it

Issue #1470, opened 2026-07-16
[Valkey] Sentinel Kubernetes master service


akoyaxd:
### Name and Version

Valkey 0.24.3

### What is the problem this feature will solve?

Valkey with sentinel high availability does not work when the Client does not support sentinel.

### Describe the solution you'd like?

Similar to the Redis Chart, which has a `sentinel.masterService.enabled` field in its helm chart: Implement a master discovery, that creates a Kubernetes service, that always points tot the current master pod.

### What alternatives have you considered?

Using the Redis Helm Chart

### Affected Helm charts

Valkey
Redis

wittfabian:
+1 on the use case — we hit exactly this. We run this chart with `architecture: replication` + Sentinel for three production apps whose clients don't speak Sentinel (node-redis v4 in a Next.js cache handler, ioredis on the auth hot path), so a chart-native stable master endpoint would be very welcome.

That said: the obvious move here is to port `sentinel.masterService` from the Redis chart 1:1, and we'd like to argue against that. We evaluated it as a replacement for what we run today and decided it would be a downgrade for us. Since the Valkey implementation doesn't exist yet, there's a chance to do it better rather than inherit the design.

## What we run today

A small HAProxy TCP proxy deployed via `extraObjects` in the same release, exposing a stable `valkey-master:6379`:

```
backend valkey-master-be
  mode tcp
  option tcp-check
  tcp-check connect
  tcp-check send "AUTH __VALKEY_PASSWORD__\r\n"
  tcp-check expect string +OK
  tcp-check send "PING\r\n"
  tcp-check expect string +PONG
  tcp-check send "INFO replication\r\n"
  # Only the pod answering role:master is marked healthy.
  # min-recv 100 makes replicas fail fast (<50ms) instead of hitting a check timeout.
  tcp-check expect string role:master min-recv 100
  tcp-check send "QUIT\r\n"
  tcp-check expect string +OK

  server valkey-0 valkey-0.valkey-headless.<ns>.svc.cluster.local:6379 check inter 1s rise 2 fall 2
  server valkey-1 valkey-1.valkey-headless.<ns>.svc.cluster.local:6379 check inter 1s rise 2 fall 2
  server valkey-2 valkey-2.valkey-headless.<ns>.svc.cluster.local:6379 check inter 1s rise 2 fall 2
```

Properties we specifically rely on:

[...]

PR #1523, opened 2026-07-31
[valkey] Add sentinel.masterProxy: HAProxy sidecar for a stable master endpoint

lichdencor:
### Description of the change

Valkey with Sentinel doesn't give clients a way in if they can't speak the Sentinel protocol - that's #1470. This adds `sentinel.masterProxy`, which bundles an HAProxy sidecar into every Valkey pod (next to the existing Sentinel sidecar) and gives you a `<release>-master` Service that always points at whoever the current master is.

HAProxy doesn't ask Sentinel or the Kubernetes API who the master is. It checks each Valkey pod directly with `INFO replication` every second and only forwards traffic to whichever one answers `role:master`. Sentinel keeps doing what it already does - detecting the failure, promoting a replica - HAProxy just reacts to that on the data plane, independently.

This follows up on the discussion in #1470 around not directly porting the Redis chart's `sentinel.masterService` (kubectl exec into a Sentinel pod + patch the Service selector) into Valkey. That design has some real weaknesses worth avoiding here: a 60s poll interval by default, `pods/exec` RBAC, `--force-conflicts --field-manager=helm` to fight Helm for ownership of the selector on every upgrade, and - the one that matters most - it never closes the client's existing connection to the demoted master, so a pooled connection can sit there getting `-READONLY` on writes until the app happens to notice.

What's in this PR:

- `sentinel.masterProxy.*` values, disabled by default
- an `haproxy` container added to the StatefulSet, with its own readiness/liveness probes
- a static `<release>-master` Service - the selector never changes, all 3 pods run the sidecar, whichever one you land on forwards you to the real master
- helm-unittest coverage in `tests/sentinel-master-proxy_test.yaml`
- README and values.schema.json updated

A couple of things worth calling out that aren't obvious from just reading the diff:

- HAProxy resolves each pod's hostname through a `resolvers` block with periodic re-resolution, not once at startup. Without that, a StatefulSet pod that restarts with a new IP gets stuck marked down forever, since HAProxy's default behavior is to resolve server hostnames once at boot.
- Trusting a plain `role:master` answer isn't enough on its own: a pod that restarts with empty/ephemeral storage can come back up self-declared master before Sentinel has finished failing over, so for a window you get two pods both answering `role:master` - I hit this in testing. To guard against it, the proxy runs two HAProxy backends: `valkey_master_strict` also requires an attached replica (`connected_slaves > 0`) and no failover in progress (`master_failover_state:no-failover`), and the frontend routes there via `nbsrv()` whenever a candidate qualifies, falling back to `valkey_master_any` (plain `role:master`) otherwise so the endpoint never goes fully unavailable (a fresh install or a resyncing replica can transiently leave nobody qualifying for the strict backend). This narrows the split-brain window rather than closing it outright - if two candidates satisfy the strict backend at the same instant (a replica split across the old and new master), traffic can still get round-robined between both until one drops out. Happy to discuss whether that residual risk is acceptable or whether a different mechanism is preferred.
- TLS isn't supported yet - enabling both `tls.enabled` and `sentinel.masterProxy.enabled` fails at render time with an explicit error instead of quietly producing something broken.

Tested end-to-end against a local k3d cluster (3 replicas, Sentinel enabled): force-killed the master pod under write load and the client reconnected cleanly through the proxy with no lasting split-brain, and normal operation was confirmed to route to the correct master.

### Benefits

- No RBAC, ServiceAccount, or Kubernetes API access needed at all - it's a plain TCP proxy watching the data plane.
- Reacts to a failover in about as long as the health check takes (a couple seconds), not a poll interval.
- Closes stale connections to the demoted master instead of leaving them open.
- Proxy redundancy scales with `replicaCount` automatically since it's a sidecar, not a separate Deployment with its own replica count to manage.

### Possible drawbacks

- One more container per pod (small footprint, but it's there).
- `kubectl get endpoints <release>-master` will show all 3 pods, not just the current master, since any pod's sidecar can forward to the real one. `kubectl logs <pod> -c haproxy` is the way to check instead - it logs every backend UP/DOWN transition.
- The split-brain guard narrows but doesn't fully close the window described above where two pods could momentarily both look like a valid master.
- No TLS support yet.

### Applicable issues

Addresses #1470

### Additional information

Open to adjusting based on what direction maintainers prefer. This leans towards the proxy option discussed in the issue rather than a controller that patches Sentinel's opinion into the Service, mainly because it's the only one of the two that actually closes stale connections rather than just answering the "who's master" question faster.

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to semver

### DIFF
--- a/charts/valkey/templates/statefulset.yaml
+++ b/charts/valkey/templates/statefulset.yaml
@@ -525,6 +525,99 @@ spec:
           {{- end }}
           resources: {{- toYaml .Values.sentinel.resources | nindent 12 }}
         {{- end }}
+        {{- if and .Values.sentinel.enabled .Values.sentinel.masterProxy.enabled (eq .Values.architecture "replication") }}
+        {{- if .Values.tls.enabled }}
+        {{- fail "sentinel.masterProxy is not yet supported together with tls.enabled" }}
+        {{- end }}
+        - name: haproxy
+          securityContext: {{ include "cloudpirates.renderContainerSecurityContext" . | nindent 12 }}
+          image: {{ include "valkey.masterProxy.image" . }}
+          imagePullPolicy: {{ .Values.sentinel.masterProxy.image.pullPolicy }}
+          {{- if .Values.auth.enabled }}
+          env:
+            - name: VALKEY_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "valkey.secretName" . }}
+                  key: {{ include "valkey.passwordKey" . }}
+          {{- end }}
+          volumeMounts:
+            - name: haproxy-config
+              mountPath: /tmp
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -e
+              cat > /tmp/haproxy.cfg << 'EOF'
+              global
+                  maxconn 2000
+
+              defaults
+                  mode tcp
+                  timeout connect 2s
+                  timeout client 30s
+                  timeout server 30s
+
+              resolvers cluster_dns
+                  parse-resolv-conf
+                  hold valid 10s
+                  hold other 30s
+                  hold refused 30s
+                  hold nx 30s
+                  hold timeout 30s
+                  hold obsolete 30s
+
+              frontend valkey_master_fe
+                  bind *:{{ .Values.sentinel.masterProxy.containerPort }}
+                  default_backend valkey_master_be
+
+              backend valkey_master_be
+                  mode tcp
+                  option tcp-check
+                  tcp-check connect
+                  {{- if .Values.auth.enabled }}
+                  tcp-check send "AUTH __VALKEY_PASSWORD__\r\n"
+                  tcp-check expect string +OK
+                  {{- end }}
+                  tcp-check send "PING\r\n"
+                  tcp-check expect string +PONG
+                  tcp-check send "INFO replication\r\n"
+                  tcp-check expect string role:master min-recv 100
+                  {{- if .Values.sentinel.masterProxy.requireConnectedReplica }}
+                  tcp-check expect ! string connected_slaves:0
+                  {{- end }}
+                  tcp-check send "QUIT\r\n"
+                  tcp-check expect string +OK
+                  {{- range $i := until (int $.Values.replicaCount) }}
+                  server {{ include "valkey.fullname" $ }}-{{ $i }} {{ include "valkey.fullname" $ }}-{{ $i }}.{{ include "valkey.fullname" $ }}-headless.{{ include "cloudpirates.namespace" $ }}.svc.{{ $.Values.clusterDomain }}:{{ $.Values.service.port }} check inter {{ $.Values.sentinel.masterProxy.checkInterval }} rise {{ $.Values.sentinel.masterProxy.checkRise }} fall {{ $.Values.sentinel.masterProxy.checkFall }} resolvers cluster_dns resolve-prefer ipv4 init-addr none
+                  {{- end }}
+              EOF
+              {{- if .Values.auth.enabled }}
+              sed -i "s/__VALKEY_PASSWORD__/${VALKEY_PASSWORD}/" /tmp/haproxy.cfg
+              {{- end }}
+              exec haproxy -f /tmp/haproxy.cfg -db
+          ports:
+            - name: haproxy
+              containerPort: {{ .Values.sentinel.masterProxy.containerPort }}
+              protocol: TCP
+          readinessProbe:
+            tcpSocket:
+              port: haproxy
+            initialDelaySeconds: 2
+            periodSeconds: 5
+            timeoutSeconds: 2
+            failureThreshold: 3
+          livenessProbe:
+            tcpSocket:
+              port: haproxy
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 2
+            failureThreshold: 3
+          resources:
+            {{- toYaml .Values.sentinel.masterProxy.resources | nindent 12 }}
+        {{- end }}
         {{- if .Values.metrics.enabled }}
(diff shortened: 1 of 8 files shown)


charts/valkey/values.yaml
```diff
@@ -472,6 +472,57 @@ sentinel:
...
+    image:
+      ## @param sentinel.masterProxy.image.registry HAProxy image registry
+      registry: docker.io
+      ## @param sentinel.masterProxy.image.repository HAProxy image repository
+      repository: haproxy
+      ## @param sentinel.masterProxy.image.tag HAProxy image tag
+      tag: "3.0-alpine"
+      ## @param sentinel.masterProxy.image.pullPolicy HAProxy image pull policy
+      pullPolicy: IfNotPresent
+    ## @param sentinel.masterProxy.containerPort Port the HAProxy sidecar listens on (must differ from service.port, which the Valkey container in the same pod already uses)
```
zOnlyKroks: Could this be enforced somehow?

charts/valkey/templates/statefulset.yaml
```diff
@@ -525,6 +525,125 @@ spec:
...
+                  tcp-check send "INFO replication\r\n"
+                  tcp-check expect string role:master min-recv 100
+                  tcp-check send "QUIT\r\n"
+                  tcp-check expect string +OK
+                  {{- range $i := until (int $.Values.replicaCount) }}
+                  server {{ include "valkey.fullname" $ }}-{{ $i }} {{ include "valkey.fullname" $ }}-{{ $i }}.{{ include "valkey.fullname" $ }}-headless.{{ include "cloudpirates.namespace" $ }}.svc.{{ $.Values.clusterDomain }}:{{ $.Values.service.port }} check inter {{ $.Values.sentinel.masterProxy.checkInterval }} rise {{ $.Values.sentinel.masterProxy.checkRise }} fall {{ $.Values.sentinel.masterProxy.checkFall }} resolvers cluster_dns resolve-prefer ipv4 init-addr none
+                  {{- end }}
+              EOF
+              {{- if .Values.auth.enabled }}
+              sed -i "s/__VALKEY_PASSWORD__/${VALKEY_PASSWORD}/" /tmp/haproxy.cfg
```
zOnlyKroks: Password is not escaped here.
If auth.password (or an existing secret's password) contains /, &, \, or a newline, this sed will break (bad s/// syntax).

Consider: sed -e "s/[&/\\]/\\\\&/g"

charts/valkey/templates/statefulset.yaml
```diff
@@ -525,6 +525,125 @@ spec:
...
+                  tcp-check expect string +OK
+                  {{- end }}
+                  tcp-check send "PING\r\n"
+                  tcp-check expect string +PONG
+                  tcp-check send "INFO replication\r\n"
+                  tcp-check expect string role:master min-recv 100
+                  tcp-check send "QUIT\r\n"
+                  tcp-check expect string +OK
+                  {{- range $i := until (int $.Values.replicaCount) }}
+                  server {{ include "valkey.fullname" $ }}-{{ $i }} {{ include "valkey.fullname" $ }}-{{ $i }}.{{ include "valkey.fullname" $ }}-headless.{{ include "cloudpirates.namespace" $ }}.svc.{{ $.Values.clusterDomain }}:{{ $.Values.service.port }} check inter {{ $.Values.sentinel.masterProxy.checkInterval }} rise {{ $.Values.sentinel.masterProxy.checkRise }} fall {{ $.Values.sentinel.masterProxy.checkFall }} resolvers cluster_dns resolve-prefer ipv4 init-addr none
```
zOnlyKroks: resolve-prefer ipv4 is hardcoded regardless of .Values.ipFamily

zOnlyKroks (changes requested):
Aside from these small issues, I am happy with your PR. thank you very much :)

charts/valkey/values.yaml
lichdencor: Added a `fail` at render time if `sentinel.masterProxy.containerPort` collides with `service.port`, instead of only documenting the constraint.
```diff
+        {{- if eq (int .Values.sentinel.masterProxy.containerPort) (int .Values.service.port) }}
+        {{- fail "sentinel.masterProxy.containerPort must differ from service.port" }}
+        {{- end }}
```

charts/valkey/templates/statefulset.yaml
lichdencor: Escaped `/`, `&`, and `\` in the password before it's substituted into the sed command. Verified end-to-end in a local cluster with a password containing `/` and `&`.
```diff
-              sed -i "s/__VALKEY_PASSWORD__/${VALKEY_PASSWORD}/" /tmp/haproxy.cfg
+              ESCAPED_PASSWORD=$(printf '%s' "$VALKEY_PASSWORD" | sed -e 's/\\/\\\\/g' -e 's/[/&]/\\&/g')
+              sed -i "s/__VALKEY_PASSWORD__/${ESCAPED_PASSWORD}/" /tmp/haproxy.cfg
```

charts/valkey/templates/statefulset.yaml
lichdencor: Now follows `.Values.ipFamily`: uses `resolve-prefer ipv6` when set to ipv6, `ipv4` otherwise.
```diff
-...resolvers cluster_dns resolve-prefer ipv4 init-addr none
+...resolvers cluster_dns resolve-prefer {{ if eq $.Values.ipFamily "ipv6" }}ipv6{{ else }}ipv4{{ end }} init-addr none
```
(truncated `...` since that `server` line is long — full line is in the diff)

lichdencor:
Thank you for the review. I addressed all three points in commit `1fc6258`! Feel free to take another look and let me know if there's anything else you'd like adjusted.

PR #1532, opened 2026-08-04
[redis] Mount the config volume only when defined

archy-rock3t-cloud:
### Description of the change

The `config` volume only renders when `config.content` or `config.existingConfigmap` is set, and the main container already guards its mount the same way. The init container mounted it unconditionally, so an empty `config.content` produced a volumeMount with no matching volume and the API server rejected the StatefulSet.

```
helm template rel charts/redis --set architecture=replication --set config.content=""
```

### Benefits

The chart installs with an empty `config.content`.

### Possible drawbacks

None. The mount is unchanged whenever a config source is set, which is the default.

### Checklist

- [X] All commits are signed and verified

### DIFF
--- a/charts/redis/Chart.yaml
+++ b/charts/redis/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: redis
 description: An open source, in-memory data structure store used as a database, cache, and message broker.
 type: application
-version: 0.34.4
+version: 0.34.5
 appVersion: "8.10.0"
 keywords:
   - redis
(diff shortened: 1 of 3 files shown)


zOnlyKroks:
@archy-rock3t-cloud This looks good so far, you´d need to version bump it once again though

archy-rock3t-cloud:
@zOnlyKroks updated

zOnlyKroks:
> @zOnlyKroks updated

thank you very much :)

PR #1551, opened 2026-08-05
[keycloak] Pass managementPort to the server

archy-rock3t-cloud:
### Description of the change

`keycloak.managementPort` sets the `management` container port and all three probes, but was never passed to Keycloak. The management interface stayed on its built-in default of 9000, so any other value left the probes pointing at a port nothing listens on and the pod never became ready.

```
helm template rel charts/keycloak --set keycloak.managementPort=9001
```

### Benefits

`keycloak.managementPort` moves the management interface, which matters when 9000 is taken by a sidecar or reserved by policy.

### Possible drawbacks

None. The default renders `--http-management-port=9000`, which is Keycloak's own default.

### Checklist

- [X] All commits are signed and verified

### DIFF
--- a/charts/keycloak/Chart.yaml
+++ b/charts/keycloak/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.21.19
+version: 0.21.20
 appVersion: "26.7.0"
 keywords:
   - keycloak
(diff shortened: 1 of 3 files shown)


charts/keycloak/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: keycloak
 description: Open Source Identity and Access Management Solution
 type: application
-version: 0.21.19
+version: 0.21.21
```
zOnlyKroks: I think you ment to update to 0.21.20 here

zOnlyKroks (changes requested):

charts/keycloak/Chart.yaml
archy-rock3t-cloud: yeah, updated

PR #1637, opened 2026-08-20
[zookeeper] Honor the documented probe enable flags and startupProbe

archy-rock3t-cloud:
### Description of the change

`livenessProbe.enabled`, `readinessProbe.enabled` and the six `startupProbe` parameters are documented in values.yaml but no template reads them: both probes render unconditionally and startupProbe is never rendered at all. Wired all three up, following the kafka chart.

### Benefits

Disabling a probe disables it, and a slow forming ensemble can use a startupProbe instead of dying to the liveness probe on cold start.

### Possible drawbacks

None. Probes default to enabled and startupProbe to disabled, so the default render is unchanged.

### Checklist


zOnlyKroks:
there seem to be a few unit-test failures, those would need to addressed before merging. thank you

PR #1746, opened 2026-09-18
[keycloak] Preserve database credentials Secret on upgrade

### LINKED ISSUE #1745: [keycloak] DB credentials Secret is regenerated on every upgrade and embedded-DB lookups resolve to the wrong Secret name
### Name and Version

keycloak 0.21.10

### 1. External database (MySQL/PostgreSQL/MSSQL), no `database.password` supplied

```bash
helm template kc oci://registry-1.docker.io/cloudpirates/keycloak --version 0.21.39 \
  --set database.type=mysql \
  --set postgres.enabled=false \
  --set mariadb.enabled=false \
  --set database.host=mysql.example.com \
  --show-only templates/secret.yaml | grep db-password
```

Run the command above three times. Each run produces a **different** `db-password`:

```
render 1 -> sha256=fd0096a88accc4e0...
render 2 -> sha256=77e9776ae1e7b49d...
render 3 -> sha256=e504fa4349b3c8ed...
```

The same happens on a real `helm upgrade`: the Secret `<release>-db` is rewritten with a brand-new random password, the StatefulSet consumes it via `secretKeyRef` as `KC_DB_PASSWORD`, and the next pod restart tries to authenticate against the database with the wrong password.

### 2. Embedded database lookups use the wrong Secret name

```bash
helm template kc ./charts/keycloak --namespace my-namespace \
  --set database.type=mysql --set mariadb.enabled=true \
  --show-only templates/secret.yaml 
# the template looks up "my-namespace-mariadb"

helm template kc ./charts/mariadb --namespace my-namespace --show-only templates/secret.yaml
# the actual Secret created by the subchart is "kc-mariadb"
```

`templates/secret.yaml` looks up `printf "%s-mariadb" (include "cloudpirates.namespace" .)` / `printf "%s-postgres" (...)`, but the subchart Secrets are named `<release>-mariadb` / `<release>-postgres` (`mariadb.fullname` / `postgres.fullname`). The lookup only succeeds when the release name equals the namespace.

```yaml
database:
  type: mysql
  host: mysql.example.com
  port: "3306"
  name: keycloak
  username: keycloak
  password: ""        # not present in every upgrade invocation
postgres:
  enabled: false
mariadb:
  enabled: false
```

On an upgrade where no new database password is supplied, the chart should keep the password already stored in the existing Secret (i.e. `lookup` the current Secret before falling back to a generated value), or fail with a clear error instead of silently generating a new one. The embedded-database lookups should resolve the subchart Secret by its real name (`<release>-mariadb` / `<release>-postgres`), the same way the database URL is built since #1557.

`templates/secret.yaml` renders the database password as:

```yaml
db-password: {{ (.Values.database.password | default (randAlphaNum 32)) | b64enc | quote }}
```

in the "Fallback for external databases" branch. Unlike the embedded postgres/mariadb branches, this branch performs no `lookup` of the existing Secret, so every `helm upgrade` that does not carry `database.password` writes a new random 32-character value into `<release>-db` and Keycloak then fails to connect (e.g. `Access denied for user ...` / `password authentication failed`). Because the Secret is only consumed via `secretKeyRef`, the pod keeps working until it is restarted, which makes the failure look intermittent and tied to upgrades.

Additionally, in the embedded branches, the `lookup` calls added in 39a029c6 ("[all]: Add namespaceOverride Value to all Charts") use the namespace as the Secret name (`<namespace>-mariadb` / `<namespace>-postgres`) instead of the subchart fullname, so the generated `keycloak-db` Secret is rotated on every upgrade whenever the release name differs from the namespace.

### Additional informations

- File: `charts/keycloak/templates/secret.yaml`
  - external fallback branch: `db-password: {{ (.Values.database.password | default (randAlphaNum 32)) | b64enc | quote }}`
  - lines 24 and 33: `lookup "v1" "Secret" (include "cloudpirates.namespace" .) (printf "%s-postgres" (include "cloudpirates.namespace" .))` and the mariadb equivalent.
- Regression source for the lookup: commit 39a029c6. Before it the code was `printf "%s-postgres" .Release.Name`, which matched the subchart Secret name.
- The same naming issue was already fixed for the JDBC URL in #1557 ("Build database URL from subchart fullname"); `secret.yaml` was left behind.
- `keycloak.adminPassword` has the same pattern (`| default (randAlphaNum 32)` without `lookup`), so the admin password in the main Secret is regenerated on every render too when adminPassword/existingSecret are not set.
- No helm-unittest coverage for this: `tests/` only has `mssql-db-secret_test.yaml`; there is no test asserting that an existing Secret value is preserved on upgrade.
- Suggested fix for the external branch (same idea for both lookups, using the helpers the StatefulSet already uses):

```yaml
{{- $existingDbSecret := (lookup "v1" "Secret" (include "cloudpirates.namespace" .) (printf "%s-db" (include "keycloak.fullname" .))) }}
{{- $dbPassword := "" }}
{{- if and $existingDbSecret $existingDbSecret.data (index $existingDbSecret.data "db-password") }}
{{- $dbPassword = index $existingDbSecret.data "db-password" }}
{{- else }}
{{- $dbPassword = (.Values.database.password | default (randAlphaNum 32) | b64enc) }}
{{- end }}
...
  db-password: {{ $dbPassword | quote }}
```

and for the embedded branches: `(include "mariadb.fullname" .Subcharts.mariadb)` / `(include "postgres.fullname" .Subcharts.postgres)`.
- Happy to submit a PR with the fix and unittest coverage.

### Affected Helm charts

keycloak

rafacouto:
### Description of the change

`templates/secret.yaml` regenerated the database credentials on every upgrade:

- The external database branch ("Fallback for external databases") rendered `db-password` as `(.Values.database.password | default (randAlphaNum 32))` and never looked up the existing Secret, so any upgrade that did not carry `database.password` overwrote `<release>-db` with a new random password. Keycloak then failed to authenticate against the database as soon as the pod was recreated.
- The embedded database branches looked up the subchart Secret by the release namespace (`<namespace>-mariadb` / `<namespace>-postgres`) instead of the subchart fullname, so the lookup never found the existing Secret unless the release name matched the namespace (regression from 39a029c6).
- The generated admin password had the same problem: `admin-password` was regenerated on every render.

The password already stored in the cluster is now reused, and a generated value is only used when no Secret exists yet (fresh installs). Same approach already used by the mariadb/postgres subcharts.

### Benefits

Upgrading the chart no longer rotates the credentials that Keycloak uses to reach the database or the bootstrap admin password, which removes a source of broken upgrades and manual rollbacks.

### Possible drawbacks

- Behaviour change: to deliberately rotate a generated password now, the Secret must be deleted (or the password supplied through `database.password` / `keycloak.adminPassword`) instead of relying on an upgrade.
- Known limitation (pre-existing, not addressed here): when `mariadb.auth.existingSecret` / `postgres.auth.existingSecret` are used, the StatefulSet still reads `KC_DB_PASSWORD` from the subchart Secret name.

### Applicable issues

- fixes #1745

### Additional information

Tests run:

- `helm unittest charts/keycloak --with-subchart=false` -> 9 suites, 79 tests, all passing.
- `helm lint charts/keycloak` -> clean.
- New `tests/secret_test.yaml` (5 cases): admin password from values / generated, external database password from values / generated, no Secret when `database.type` is empty.

Note: `lookup` returns an empty result under `helm template` / `helm unittest`, so the unit tests only exercise the fresh-install path; the preservation path requires a live cluster (`helm install` + `helm upgrade`) to be exercised.

### Checklist

- [X] Variables are documented in the values.yaml and added to the `README.md` (not applicable: no new values introduced)


zOnlyKroks (dismissed):
If mariadb.auth.username gets set on an upgrade where it wasn't set before, $mariadbPasswordKey switches to the user-password key. The live secret from mariadb's subchart may only have the root-password key populated in that case. index $existingMariadbSecret.data $mariadbPasswordKey returns nil, and the templated value renders as an empty string. Producing an invalid db-password: entry with no value.

